PR #12, opened 2021-03-08
Revert "Started working on test case"

aujjwal-redhat:
Reverts aujjwal-redhat/RedAnt_Test#11


tests/test_peer_probe.py
```diff
@@ -1,31 +0,0 @@
...
-print(sys.path)
-from redant_libs.peer_ops import (
-    peer_probe
-)
-
-from odinControl.rexe.rexe import Rexe
-
-from redant_libs.redant_resources import Redant_Resources as RR
-from unittest import TestCase
-import unittest
```
srijan-sivakumar, line 13: The idea is not to use unittest or pytest

tests/test_peer_probe.py
```diff
@@ -1,31 +0,0 @@
...
-
-R = Rexe(conf_path="./Utilities/conf.yaml")
```
srijan-sivakumar, line 15: A test case should only be concerned with the steps to be run. Not other configurations.

tests/test_peer_probe.py
```diff
@@ -1,31 +0,0 @@
...
-
-        R.establish_connection()
-        RR.rlogger.info("Started peer probe testing")
-        ret = peer_probe('10.70.43.101','10.70.43.228')
-
-        self.assertEqual(ret, 0, "Peer probe failed")
-        RR.rlogger.info("Ended peer probe testing")
-
-
-if __name__ == '__main__':
```
srijan-sivakumar, line 29: Again. not required.

tests/test_peer_probe.py
```diff
@@ -1,31 +0,0 @@
...
-
-class TestPeerProbe(TestCase):
-
-    def test_peer_probe(self):
-
-        R.establish_connection()
-        RR.rlogger.info("Started peer probe testing")
-        ret = peer_probe('10.70.43.101','10.70.43.228')
-
-        self.assertEqual(ret, 0, "Peer probe failed")
```
srijan-sivakumar, line 25: This goes against the design. Please go through the design documentation and see as to how it was envisaged.

tests/test_peer_probe.py
```diff
@@ -1,31 +0,0 @@
...
-
-from odinControl.rexe.rexe import Rexe
-
-from redant_libs.redant_resources import Redant_Resources as RR
-from unittest import TestCase
-import unittest
-
-R = Rexe(conf_path="./Utilities/conf.yaml")
-
-class TestPeerProbe(TestCase):
```
srijan-sivakumar, line 17: I think we won't be needing the use of a class wrt our design. The tests are parsed and after their ordering has been done, the test runner will take a TC from the list, decide what all servers and client will be required and pass this machine list and the TC to the thread runner. The thread runner will then create the connections based on the given configuration from the test runner. Once the connections are done, the TC which inturn contains a function will be invoked by the runner thread. The test steps will be run. Instead of assertions we will have `try-except` clause.

srijan-sivakumar:
@aujjwal-redhat , @neesingh-rh , @nishith-vihar  Please go through the review comments before working on the TC. Also, go through the design doc. I'd suggest that you also add Barak for reviews and do not merge the code before getting a review flag for a go ahead.

PR #13, opened 2021-03-08
Volume mount operation is added

nishith-vihar:


redant_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
...
+    node = "10.70.43.63"
+
+    if force:
+        cmd = "mount -t --force glusterfs %s:/%s %s" % (node,volname,path)
+    else:
+        cmd = "mount -t glusterfs %s:/%s %s" % (node,volname,path)
+
+    ret = R.execute_command(node=mnode,cmd=cmd)
+
+    #TODO: to be removed later
```
srijan-sivakumar, line 33: What is this todo for ? If it is for validation, we can start working on it rather than leaving that as debt for later.

nishith-vihar: Actually the ip address for the node has to come from config file I guess. The parsing module is not yet connected to ops. So I have just hard-coded the ip for testing.

srijan-sivakumar: ok, but this seems to be for the print

redant_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
...
+        force (bool): If this option is set to True, then mount volume
+            will get executed with force option. If it is set to False,
+            then mount volume will get executed without force option
+    """
+
+
+    RR.rlogger.info("Volume Mount Command initiated")
+    
+    #TODO: 'node' variable should be removed later
+    node = "10.70.43.63"
```
srijan-sivakumar, line 24: We can pass this as an argument rather than hardcoding in here the function.

srijan-sivakumar: I'd recommend not to create this debt in here as we'd just keep on piling such additions across the code. Focus on adding the parser into the code-base, write a sample TC even if it is for a couple of lines.

srijan-sivakumar (changes requested):

redant_libs/volume_ops.py
nishith-vihar: yeah. Actually the print is for testing the return value which has to be removed.

redant_libs/gluster_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
 #TODO: this is for testing. When test runner will be ready it will be removed.
 R = Rexe(conf_path="./example/conf.yaml")
```
srijan-sivakumar, line 5: Not required in the ops libraries from now on.

redant_libs/gluster_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
...
 
+def volume_mount(mnode , volname , path,force=False):
```
srijan-sivakumar, line 7: use the function annotations.

redant_libs/gluster_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
...
+        path(str): path on client node to be mounted onto volume
+        force
+    Kwargs:
+        force (bool): If this option is set to True, then mount volume
+            will get executed with force option. If it is set to False,
+            then mount volume will get executed without force option
+    """
+
+
+    RR.rlogger.info("Volume Mount Command initiated")
```
srijan-sivakumar, line 21: Refer the peer_ops for reference on the usage of logging and call to remote command executor.

redant_libs/gluster_libs/volume_ops.py
```diff
@@ -4,6 +4,39 @@
...
+    
+    #TODO: 'node' variable should be removed later
+    node = "10.70.43.63"
```
srijan-sivakumar, line 24: Remove this hard coded value from code. To test the functions, we should only create small TCs. Not to be done here directly.

srijan-sivakumar (changes requested):
Do refer to the latest design considerations. I'll add a readme.md stating python coding considerations, but those can go as a separate PR but do check out the peer_ops on how an ops lib is re-deisnged.

redant_libs/gluster_libs/volume_ops.py
nishith-vihar: Please refer to #24

PR #18, opened 2021-03-09
Modified peer_ops and redant_resources

aujjwal-redhat:
Modified the import paths in redant_resources and peer_ops so that it can start running


srijan-sivakumar (approved):
I'm approving this PR, but we need to see how a TC can send across the remote_executor object's reference to the ops libraries, rather than creating the Rexe in the library ops;

PR #24, opened 2021-03-10
Exceptions and Logging

nishith-vihar:
Exceptions are raised based on the return value and logging
is done accordingly complying to the design doc.

### DIFF
--- a/redant_libs/gluster_libs/volume_ops.py
+++ b/redant_libs/gluster_libs/volume_ops.py
@@ -81,8 +81,6 @@ def volume_create(mnode,volname,bricks_list,force=False, **kwargs):
                               disperse, disperse_data, redundancy,
                               transport, ' '.join(bricks_list)))
     
-    RR.rlogger.debug("Creating gluster volume...")
-    
     if force:
         cmd = cmd + " force"
 
@@ -94,8 +92,6 @@ def volume_create(mnode,volname,bricks_list,force=False, **kwargs):
     if int(ret['msg']['opRet'])!=0:
         RR.rlogger.error(ret['msg']['opErrstr'])
         raise Exception(ret['msg']['opErrstr'])
-    
-    RR.rlogger.debug("Volume created successfully.")
 
 
 def volume_start(mnode,volname,force=False):
@@ -109,8 +105,6 @@ def volume_start(mnode,volname,force=False):
             then start volume will get executed without force option
     Logging is done and exceptions are raised if required
     """
-
-    RR.rlogger.debug("Starting gluster volume...")
 
     if force:
         cmd = "gluster volume start %s force --mode=script --xml" % volname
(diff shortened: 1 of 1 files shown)


redant_libs/gluster_libs/volume_ops.py
```diff
@@ -92,9 +91,13 @@ def volume_create(mnode,volname,bricks_list,force=False, **kwargs):
...
     print(ret)
     
-    RR.rlogger.info(ret)
+    if int(ret['msg']['opRet'])!=0:
+        RR.rlogger.error(ret['msg']['opErrstr'])
+        raise Exception(ret['msg']['opErrstr'])
     
-    return ret
+    RR.rlogger.info("Volume created successfully.")
+    RR.rlogger.debug(ret)
```
nik-redhat, line 99: Any use case for adding these?

nishith-vihar: Nothing specific. just appropriate logging to make it easier during debugging if we encounter an error.

nik-redhat: Nope, not the success message but the debug log. It can be more elaborated to be clear to the user.
Does it make sense or just overthinking @srijan-sivakumar @BarakSason ?

srijan-sivakumar: I agree with Nikhil, let's keep the success case logging for Debug level instead of info as that'll just add too many log outputs in the logfile.

srijan-sivakumar: As per the info log added, I've always thought it to be a little painful to read if there are multiple operations in a TC. I think we can maybe handle it from the Result side itself..if any of the steps failed, we would come to know as to why it failed and where..if it succeeded, I think we can just say success rather than putting an extra log output for every function which executed successfully...

nik-redhat:
If the success logs are DEBUG level, then shouldn't the creation logs also be at the same level?

BarakSason:
Logging is extremely important aspect of the framework.

IMO the best way to handle logging in th ops libraries it the follow the following format:
... op_command(args) {
<Info level log describing the op and which machines the op is going to be performed on>
<execute op>

on success - <info level log describing the op is successful and which machines the op is going to be performed on>
on failure - <construct failure string, describing what failed and on what machines and throw an exception containing the info to be caught in test code>
}

As we discussed yesterday, I see little value in DEBUG log level. We're dealing with a test framework, for the vast majority (if not all) cases, INFO and ERROR levels are sufficient.

One additional request - Please modify the logger code to only include a single API:
logger.log(string, LOG LEVEL)
The number of APIs should be kept to a minimum, and in this case we can achieve the same result with only a single API and pass the log level as an argument.

srijan-sivakumar:
I feel we  might need other log levels. Though not visible right now, but debug logs actually comes handy. Also pushing everything to info ( in success scenarios ) will leave us in the same state as seen in glusto.

srijan-sivakumar:
I presume, @aujjwal-redhat took the inspiration from odinControl for the logging part, which is just a simple logger.

@BarakSason I think we should have a discussion over the logging part to make it more mature. I feel it's more on how to log rather than the logger module.

srijan-sivakumar (approved):

PR #25, opened 2021-03-10
Added format specifiers

aujjwal-redhat:
Added format specifiers in peer_ops.


nik-redhat (commented):
Keep some breathing room at least.....code should be written in a way so there is less rework needed later on.

aujjwal-redhat:
Is it fine now?

PR #30, opened 2021-03-11
Make logging API a little generic

srijan-sivakumar:
Changes:
1. Added a new method `rlog` to handle the generic way
of calls to logging. It takes two arguments of type string,
log_message and log_level wherein the default log_level is Info.
2. Modified the thread_runner and rexe with the logging changes.
3. Added logging.md to put out the design considerations.

### DIFF
--- a/redant_libs/support_libs/logging.md
+++ b/redant_libs/support_libs/logging.md
@@ -1,0 +1,41 @@
+# Redant Logging.
+
+Redant logging is achieved with the help of the `Logger` class which currently
+consists of three methods,
+1. set_logging_options
+2. rlog
+3. get_logger_handle
+
+Wherein the set_logging_options and get_logger_handle are static methods while
+rlog is a class method. The current design is a little convoluted as it exposes
+the logger object to the outside world which has to be rectified. If one looks
+into the class we have a `has a` relation between our logger object and the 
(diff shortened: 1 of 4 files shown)


aujjwal-redhat (approved):
LGTM.

redant_libs/support_libs/logging.md
```diff
@@ -0,0 +1,41 @@
...
+parameter i.e. the log-level.
+2. The log levels to be used by the user while invoking the methods are as follows,
+
+	| Log level | Parameter |
+	|:---------:|:---------:|
+	|   Debug   |    'D'	|
+	|   Info    |    'I'	|
+	|   Error   |    'E'    |
+	|   Warning |    'W'    |
+        |   Critical |   'C'    |
```
nik-redhat, line 29: We agreed upon having only levels, right?

srijan-sivakumar: This is just a standard log level, the guideline would reflect the usage of info and error logs.

nik-redhat: But, then someone might find it confusing if we allow it?
As I see we are configuring the same in logger class as well. It can be bypassed, accidentally.

Issue #32, opened 2021-03-11
Mixin Pattern across the base classes


srijan-sivakumar:
The current dependency flow is a little circuitous and can lead to issues. I'm proposing we combine all the base support libraries and ops into a framework_mixin which can be used across the test functions instead of individual calls to the classes and modules.

PR #33, opened 2021-03-11
Mixin pattern across the base classes

srijan-sivakumar:
Code changes:
1. Added redant_mixin to be invoked by the thread runner
and has been integrated in case of rexe and relog.
2. Redant_mixin design thought process has been added
as a markdown.
3. naming conventions changed across the thread_runner

What next:
1. Other ops libraries need to follow this template.

Fixes: #32


redant_libs/gluster_libs/peer_ops.py
```diff
@@ -1,172 +1,91 @@
...
-        else:
-            cmd = "gluster --xml peer detach %s --mode=script" % server
-
-        ret = R.execute_command(node, cmd)
-
-        #TODO: to be removed
-        pp.pprint(ret)
+class peer_ops:
+
+    def peer_probe(server, node):
```
nik-redhat, line 3: Let's switch that to keep it aligned with the glusto function.

srijan-sivakumar: I'm not checking the function names and class names. I just wanted to create a reference to be taken up for the ops and support libraries to be added. But yes, better to have a similar nomenclature as that of glusto libraries for easier conversion of TCs in future.

nik-redhat: There are other places as well, which need to be tweaked a bit. Like, using the same `name` for a particular reference instead of switching them.
But, again that can be done in a separate PR.

srijan-sivakumar: Ack. That should be addressed as a separate issue in my opinion. This PR is already big enough. Better not to repeat the `that` and `THAT` scenario :p

nik-redhat:
This doesn't handle the issue of `same name` functions, or does it?

srijan-sivakumar:
> This doesn't handle the issue of `same name` functions, or does it?

Technically I'm not sure if such case arises in mixin if it'd throw an error or take the very first class and execute it. But statistically speaking we won't be having functions with same name across our project. And for that, I leave it to the linting tools to pick it up if there's some issue.

nik-redhat:
> > This doesn't handle the issue of `same name` functions, or does it?
> 
> Technically I'm not sure if such case arises in mixin if it'd throw an error or take the very first class and execute it. But statistically speaking we won't be having functions with same name across our project. And for that, I leave it to the linting tools to pick it up if there's some issue.

Ack...points mentioned https://www.ianlewis.org/en/mixins-and-python here are a bit helpful in configuring that.

nik-redhat (approved):

PR #35, opened 2021-03-11
Added try-except and root dir in func

neesingh-rh:


srijan-sivakumar:
@neesingh-rh , do add description and also, better to link this with an issue. So, before adding a feature or fixing a bug, create an issue and then while committing the code, do `git commit -s`. 
1. Add a small header detail. ( probably same as that of issue heading )
2. Add the details about what this commit is about.
3. Add whether this patch fixes the issue by adding a line `Fixes #<issue_number>` or `Updates #<issue_number>`
Please follow this methodology as it would be easy for reviewers to review the code and also to see as to what is being done in here.

It also comes handy for posterity.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,12 +1,9 @@
 import sys
+import os
 
-sys.path.append("./odinControl")
+from redant_libs.support_libs.rexe import Rexe
```
srijan-sivakumar, line 4: I have added a PR to handle these imports. Refer to #33

redant_libs/support_libs/io_ops.py
```diff
@@ -33,59 +30,88 @@
...
+            raise Exception(ret['msg'])
+    
+    except Exception as e:
+        RR.rlogger.error(e)
+
+    
+    return ret
 
 
 if __name__ == "__main__":
```
srijan-sivakumar, line 111: I'd recommend writing a simple TC in tests to verify these functions instead of doing it here.

redant_libs/support_libs/io_ops.py
```diff
@@ -33,59 +30,88 @@
...
-    RR.rlogger.info(ret)
-    return ret
+        RR.rlogger.info(ret)
 
+        if ret['error_code'] != 0:
+            raise Exception(ret['msg'])
+    
+    except Exception as e:
+        RR.rlogger.error(e)
+
```
srijan-sivakumar, line 106: unnecessary space.

redant_libs/support_libs/io_ops.py
```diff
@@ -33,59 +30,88 @@
...
     
-    cmd = 'ls /root_dir'
+        cmd = 'ls /'+ ROOT_DIR
 
-    ret = R.execute_command(node="192.168.122.161", cmd=cmd)
+        ret = R.execute_command(node="192.168.122.161", cmd=cmd)
 
-    #TODO: to be removed
-    pp.pprint(ret)
+        #TODO: to be removed
```
srijan-sivakumar, line 96: please remove the pprints. Logger should be the only point of communication as to what is happening in a TC.

srijan-sivakumar (changes requested):

redant_libs/support_libs/io_ops.py
neesingh-rh: Okay,I will do that

Issue #42, opened 2021-03-12
Logging messages to follow the required format


aujjwal-redhat:
### Log Message format
===================
All log messages MUST include:
* The cmd that is/was executed
* The machine(s) on which the command is/was executed

Ex: peer_ops

On starting the execution of cmd on a node:
```js
 self.rlog("Running %s on node %s" % (cmd,node), 'I')
```
On completing the execution:
```js
 self.rlog("Successfully ran %s on %s " % (cmd, node),'I')
```
### Log levels
===========

* Everything in the ops library has to be in info mode.
* Everything in the test cases has to be in error mode.
* Everything other than ops to be in debug mode.

aujjwal-redhat:
@srijan-sivakumar @BarakSason Do verify it, please.

srijan-sivakumar:
@aujjwal-redhat , this design decision can be added in the logging.md so that it is documented.

aujjwal-redhat:
Ok

srijan-sivakumar:
Let's keep this open till the relevant patches are merged.

Issue #31, opened 2021-03-11
Following pep-3107 standard


srijan-sivakumar:
We all know the python is slick in handling types, for instance we can create a function called

```
def sample_func(sample_var):
   print(sample_var)
```

and this function can take in any data type and print it ( even functions as functions are first class objects in python ). So this can be disastrous for us who are designing the framework as this can lead to issues when somebody invokes and uses the function
in a way which isn't intended. The question is, do we have a salvation ?

Yes,

PEP-3107 : https://www.python.org/dev/peps/pep-3107/

The very least we can do is annotate the parameters so that it is seen in the function definitions when the user if going through them. If anyone uses IDEs, these annotations would help in correcting a user if they're going to commit a mistake in the type of data they're going to pass.

The relog module already uses function annotations, and can be used as a reference for further changes across the framework.

srijan-sivakumar:
Converting this into a discussion as I think the standards are being followed in the code base.

PR #45, opened 2021-03-12
Changes acc to format

neesingh-rh:
-> Added Try-Except blocks
-> Removed the extra spaces and the imports
-> Created the class and defined all ops func inside it.
-> Used format specifiers

### DIFF
--- a/redant_libs/support_libs/io_ops.py
+++ b/redant_libs/support_libs/io_ops.py
@@ -1,117 +1,77 @@
 import sys
 import os
 
-from redant_libs.support_libs.rexe import Rexe
+class io_ops:
 
-from redant_libs.redant_resources import Redant_Resources as RR
+    def volume_mount(self ,node , server , volname , dir,force=False):
+        try:
+            self.log("Volume Mount Command initiated")
+            
+            if force:
+                cmd = "mount -t --force glusterfs %s:/%s /%s" %s (server,volname,dir)
+            else:
+                cmd = "mount -t glusterfs %s:/%s /%s" %s (server,volname,dir)
+            ret = self.execute_command(node=node,cmd=cmd)    
+            
+            self.log(ret)    
+        
+            if ret['error_code'] != 0:
+                raise Exception(ret['msg'])
+        
+        except Exception as e:
+            self.rlog(e) 
+        return ret 
 
-import pprint   # to print the output in a better way and hence more understandable
+    def touch(self, file_name , node):
 
-#TODO: test runner thread will provide the path. Using the below object temporarily
-R = Rexe(conf_path="./Utilities/conf.yaml")
+        try:
+            """Creates a regular empty file"""
+            self.rlog("Creating File") 
 
-pp = pprint.PrettyPrinter(indent=4)
+            cmd = "touch {}".format(file_name)
+            ret = self.execute_command(node=node, cmd=cmd)
 
+            self.rlog(ret)
 
-""" def volume_mount(mnode , volname , dir,force=False):
-    RR.rlogger.info("Volume Mount Command initiated")
+            if ret['error_code'] != 0:
+                raise Exception(ret['msg'])
+        
+        except Exception as e:
+            self.rlog(e)    
+        return ret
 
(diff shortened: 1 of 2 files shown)


redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
-    create_dir("test_dir")
-    list_files_on_root_dir()
-   
+import os
+
+class io_ops:
+
+    def volume_mount(self ,node , server , volname , dir,force=False):
+        try:
+            self.log("Volume Mount Command initiated")
```
srijan-sivakumar, line 8: Do check the logging function and also follow the logging design considerations. Please refer to : https://github.com/srijan-sivakumar/Redant/issues/42

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
+            self.log(ret)    
+        
+            if ret['error_code'] != 0:
+                raise Exception(ret['msg']['opErrstr'])
+        
+        except Exception as e:
+            self.rlog(e) 
+        return ret 
+
+    def touch(self, file_name , node):
```
srijan-sivakumar, line 25: What does this function do? Do add function docstring. Also, we are to follow the function annotations from now on. Do check the issue : https://github.com/srijan-sivakumar/Redant/issues/31

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
+
+        try:
+            """Creates a regular empty file"""
+            self.rlog("Creating File")
```
srijan-sivakumar, line 29: The logging will not just say what is being done. It would encompass as to what cmd is to be run and in which server. Do check the issue : https://github.com/srijan-sivakumar/Redant/issues/42

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
+        except Exception as e:
+            self.rlog(e)    
+        return ret
+
+    def mkdir(self,dir_name, node):
+
+        try:
+            self.rlog("Creating Directory")
+        
+            ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
srijan-sivakumar, line 48: The mkdir will be creating a directory in a remote server. In this scenario we are taking the absolute path from the execution server which is a different system. Instead of dirname, we should be taking in the dir_path.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
+        except Exception as e:
+            self.rlog(e)    
+        return ret
+
+    def ls(self , node):
+
+        try:
+            self.rlog("List the files on root directory")
+
+            ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
srijan-sivakumar, line 66: Same as the mkdir.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,77 @@
...
-
-
-if __name__ == "__main__":
-    R.establish_connection()
-    #volume_mount("10.70.43.228","test-vol","test_dir")
-    create_file_using_touch("test_file")
-    create_dir("test_dir")
-    list_files_on_root_dir()
-   
+import os
```
srijan-sivakumar, line 2: Won't be required as mkdir and ls are using this but the current logic takes remote path in the system wherein the Redant is running rather than the remote server. We should be taking complete paths rather than just the name.

srijan-sivakumar (changes requested):

redant_libs/support_libs/io_ops.py
neesingh-rh: my bad , it should be rlog ..I will correct it

nik-redhat: Add docstrings in each function definition describing the arguments it takes and what the function returns.

nik-redhat (changes requested):

PR #48, opened 2021-03-12
Added logging format in logging.md

aujjwal-redhat:
Modified the logging.md file:
Added format
Added Levels

Updates: #42 

### DIFF
--- a/redant_libs/support_libs/logging.md
+++ b/redant_libs/support_libs/logging.md
@@ -36,6 +36,30 @@ this seems to be a little less generic, so `rlog` was created to handle this.
  The internal mapping to the logger's API calls will be done by rlog with the help of
 a dictionary.
 
+# Logging format:
+
+All log messages MUST include:
+* The cmd that is/was executed
+* The machine(s) on which the command is/was executed
+
+Ex: peer_ops
+
+On starting the execution of cmd on a node:
+```js
+ self.rlog("Running %s on node %s" % (cmd,node), 'I')
+```
+On completing the execution:
+```js
+ self.rlog("Successfully ran %s on %s " % (cmd, node),'I')
+```
+## Log levels
+=============
+
+* Everything in the ops library has to be in info mode.
+* Everything in the test cases has to be in error mode.
+* Everything other than ops to be in debug mode.
+
+
 # TODO:
 1. Making the logger more object oriented. Re-think the current design of how
 the class is.
--- a/redant_libs/support_libs/logging.md
+++ b/redant_libs/support_libs/logging.md
@@ -51,18 +51,18 @@ self.rlog(msg, log level)
 For *INFO* level you can omit the level as it is by default **INFO**<br>
 On starting the execution of cmd on a node:<br>
 ```js
- self.rlog("Running %s on node %s" % (cmd,node))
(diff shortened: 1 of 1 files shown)


redant_libs/support_libs/logging.md
```diff
@@ -36,6 +36,30 @@ this seems to be a little less generic, so `rlog` was created to handle this.
...
+
+All log messages MUST include:
+* The cmd that is/was executed
+* The machine(s) on which the command is/was executed
+
+Ex: peer_ops
+
+On starting the execution of cmd on a node:
+```js
+ self.rlog("Running %s on node %s" % (cmd,node), 'I')
```
srijan-sivakumar, line 49: Default log level is Info so best practice is not to pass it :p

redant_libs/support_libs/logging.md
```diff
@@ -36,6 +36,30 @@ this seems to be a little less generic, so `rlog` was created to handle this.
...
+```
+On completing the execution:
+```js
+ self.rlog("Successfully ran %s on %s " % (cmd, node),'I')
```
srijan-sivakumar, line 53: How about we modify it into a more pythonic way, 
  `self.rlog(f"Successfully ran {cmd} on {node}")`.

srijan-sivakumar: I think this can be reflected all over the redant as using the formatted string literals. https://docs.python.org/3.6/reference/lexical_analysis.html#formatted-string-literals

srijan-sivakumar (changes requested):

redant_libs/support_libs/logging.md
aujjwal-redhat: This is just for a reference about the format. Though I will put something else to show.

aujjwal-redhat: Moreover, this was intended to show the message "Runnig the `cmd` on `node`"

redant_libs/support_libs/logging.md
```diff
@@ -36,6 +36,35 @@ this seems to be a little less generic, so `rlog` was created to handle this.
...
+Ex: 
+
+```js
+self.rlog(msg, log level)
+
+```
+For *INFO* level you can omit the level as it is by default **INFO**<br>
+On starting the execution of cmd on a node:<br>
+```js
+ self.rlog("Running %s on node %s" % (cmd,node))
```
srijan-sivakumar, line 54: use the formatted string. But let's take it for the next PR. Let's not clutter this more :p

redant_libs/support_libs/logging.md
```diff
@@ -36,6 +36,35 @@ this seems to be a little less generic, so `rlog` was created to handle this.
...
+On completing the execution:
+```js
+ self.rlog("Successfully ran %s on %s " % (cmd, node))
+```
+## Log levels
+=============
+
+* Everything in the ops library has to be in info mode.
+* Everything in the test cases has to be in error mode.
+* Everything other than ops to be in debug mode.
```
srijan-sivakumar, line 65: debug is present only in the remote command executioner. We can push that point directly instead of saying other than ops.

srijan-sivakumar (changes requested):

srijan-sivakumar (approved):

Issue #50, opened 2021-03-12
Using formatted string literals.


srijan-sivakumar:
The usual way of formatting string in python had been
`("sample %s" %(val) )` or `("sample %s", val)`

But the cleaner way of doing it is using the formatted strings.
`(f"sample {val}")`. I propose we use this format across the project rather than having multiple different ways of formatting our string.

For reference : https://docs.python.org/3.6/reference/lexical_analysis.html#formatted-string-literals

nik-redhat:
I was just going to open a discussion on the same topic :p . I agree with this, and let's follow one approach, whichever is cleaner.

srijan-sivakumar:
Yes..The formatted string approach reduces the code length and is also not a sore sight for eyes as compared to other formats. One can easily understand as to what would be logged easily.

aujjwal-redhat:
Yeah let's make a standards.md file and put these format rules. It's hard to just go through the issues.

nik-redhat:
> I think we need to document these python code level considerations in a doc as standards. Just issues might be hard to refer to..

+1

PR #52, opened 2021-03-12
Strings and annotations corrected

nishith-vihar:
Strings are formatted and annotations are updated
Updates: #50  , #31

### DIFF
--- a/redant_libs/gluster_libs/volume_ops.py
+++ b/redant_libs/gluster_libs/volume_ops.py
@@ -1,6 +1,8 @@
 class VolumeOps:
 
-    def volume_create(self,node,volname,bricks_list,force=False, **kwargs):
+    def volume_create(self,node: str,volname: str,
+                      bricks_list: list,force: bool=False,
+                      **kwargs):
         """Create the gluster volume with specified configuration
         Args:
             node(str): server on which command has to be executed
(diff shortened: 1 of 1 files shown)


redant_libs/gluster_libs/volume_ops.py
```diff
@@ -51,46 +53,47 @@ def volume_create(self,node,volname,bricks_list,force=False, **kwargs):
...
+        cmd = (f"gluster volume create {volname} {replica} "
+               f"{arbiter} {stripe} {disperse} {disperse_data} "
+               f"{redundancy} {transport} {bricks_list_string} --xml "
+               "--mode=script")
         
         if force:
             cmd = cmd + " force"
         
-        self.rlog("Running %s on node %s" % (cmd,node), 'I')
+        self.rlog(f"Running {cmd} on node {node}", 'I')
```
srijan-sivakumar, line 85: Do take care of passing extra values from next time..No need to pass Info flag as it is default. I'll approve it this time. But the next run when you're working on this file, do fix this..

srijan-sivakumar (approved):

redant_libs/gluster_libs/volume_ops.py
nishith-vihar: I have signed it using -s only actually

PR #54, opened 2021-03-12
Issue51

srijan-sivakumar:
This was part 2 of issue 51 :p


nik-redhat:
Safe to mark as `draft` :joy:

srijan-sivakumar:
Yup..I forgot to do that :p

redant_libs/redant_mixin.py
```diff
@@ -1,6 +1,15 @@
+"""
+    Module Name:
```
nik-redhat, line 2: Is this needed, or did you miss to add text for it?

srijan-sivakumar: I kept it for future...Need to think on what can be part of the module doc string..

nik-redhat: Ack.

PR #55, opened 2021-03-12
Annotations and format specifiers

neesingh-rh:
Added the annotations and changed the way of using format specifiers
Update: #45

### DIFF
--- a/redant_libs/support_libs/io_ops.py
+++ b/redant_libs/support_libs/io_ops.py
@@ -16,7 +16,7 @@ class io_ops:
             self.log(ret)    
         
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e) 
@@ -34,7 +34,7 @@ class io_ops:
             self.rlog(ret)
 
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e)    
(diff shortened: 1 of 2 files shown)


redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,97 @@
...
-if __name__ == "__main__":
-    R.establish_connection()
-    #volume_mount("10.70.43.228","test-vol","test_dir")
-    create_file_using_touch("test_file")
-    create_dir("test_dir")
-    list_files_on_root_dir()
-   
+class io_ops:
+
+    def volume_mount(self ,node :str ,server :str ,volname :str, dir :str ,force : bool=False):
```
nik-redhat, line 3: What does this do?

nik-redhat: Do we want this @srijan-sivakumar ?

nik-redhat:
Also, it's a good practice to `sign-off` your commits.

redant_libs/support_libs/io_ops.py
srijan-sivakumar: Are you referring to function annotations ?

srijan-sivakumar: @nik-redhat ^^

srijan-sivakumar (approved):

redant_libs/support_libs/io_ops.py
nik-redhat: Yes, what does the `:` do in the argument?

neesingh-rh: This is the type annotation used in python to let others know about the type of argument passed.

srijan-sivakumar: Ahh, it follows the same syntax as golang without the actual type checking, it is basically saying node is of type str ( But funny part is that python interpreter still takes in any value you give it :/ ) But if you work on IDE...that's where this comes handy plus anyone reading your code..

PR #64, opened 2021-03-15
Non standard func

neesingh-rh:
-> Added function for non-standard commands.

### DIFF
--- a/redant_libs/support_libs/io_ops.py
+++ b/redant_libs/support_libs/io_ops.py
@@ -16,7 +16,7 @@ class io_ops:
             self.log(ret)    
         
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e) 
@@ -34,7 +34,7 @@ class io_ops:
             self.rlog(ret)
 
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e)    
@@ -52,7 +52,7 @@ class io_ops:
             self.rlog(ret)
 
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e)    
@@ -70,7 +70,7 @@ class io_ops:
             self.rlog(ret)
 
             if ret['error_code'] != 0:
-                raise Exception(ret['msg'])
+                raise Exception(ret['msg']['opErrstr'])
         
         except Exception as e:
             self.rlog(e)    
--- a/redant_libs/support_libs/io_ops.py
+++ b/redant_libs/support_libs/io_ops.py
@@ -1,117 +1,77 @@
 import sys
 import os
 
-from redant_libs.support_libs.rexe import Rexe
+class io_ops:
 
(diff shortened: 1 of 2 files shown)


redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,118 @@
...
+        dir: The path of the mount directory(mount point)
+        '''
 
-from rexe.rexe import Rexe
+        try:
+            self.rlog("Volume Mount Command initiated", "I")
 
-from redant_resources import Redant_Resources as RR
+            if force:
+                cmd = f"mount -t --force glusterfs {server}:/{volname} /{dir}"
```
nik-redhat, line 16: Are you sure there is a `--force` flag?

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,118 @@
...
-print(sys.path,"\n\n\n")
+        node: The node in the cluster where volume mount is to be run
+        server: HOST_NAME or IP address
+        volname: Name of volume to be mounted
+        dir: The path of the mount directory(mount point)
+        '''
 
-from rexe.rexe import Rexe
+        try:
+            self.rlog("Volume Mount Command initiated", "I")
```
nik-redhat, line 13: These logs are not needed.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,118 @@
-import sys
+class io_ops:
 
-sys.path.append("./odinControl")
+    def volume_mount(self, node: str, server: str, volname: str, dir: str, force: bool = False):
+        '''Mounts the gluster volumes
 
-print(sys.path,"\n\n\n")
+        node: The node in the cluster where volume mount is to be run
+        server: HOST_NAME or IP address
```
nik-redhat, line 7: Any special reason for defining hostname in this way? It could simply be `Hostname`, no need to capitalize or separate with an underscore.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,118 @@
-import sys
+class io_ops:
 
-sys.path.append("./odinControl")
+    def volume_mount(self, node: str, server: str, volname: str, dir: str, force: bool = False):
```
nik-redhat, line 3: Add docstring for the `optional` force argument as well, in all the functions.

redant_libs/support_libs/io_ops.py
```diff
@@ -1,91 +1,118 @@
...
-from rexe.rexe import Rexe
+        try:
+            self.rlog("Volume Mount Command initiated", "I")
 
-from redant_resources import Redant_Resources as RR
+            if force:
+                cmd = f"mount -t --force glusterfs {server}:/{volname} /{dir}"
+            else:
+                cmd = f"mount -t glusterfs {server}:/{volname} /{dir}"
+            self.rlog(f"Running {cmd} on node {node}", 'I')
```
nik-redhat, line 19: `I` is the default log level so no need to pass it explicitly.

neesingh-rh: Yeah..I will remove.

redant_libs/support_libs/io_ops.py
neesingh-rh: Okay

redant_libs/support_libs/io_ops.py
neesingh-rh: Will check

Issue #70, opened 2021-03-15
Adding the parsing module


nishith-vihar:
The parsing module consists of 3 components:
1) The detailed config file
2) The Parser class which parses the config file and generates config hashmap given the path of config file
3) The ParamsHandler which parses the configuration parameters from config hashmap

The gluster_test_main file makes call to parsing module whose function is mentioned in the Gluster-test-design document.

PR #71, opened 2021-03-15
Adding the Parsing module

nishith-vihar:
Containes Parser and ParamsHandler along with config file.
gluster_test_main is also added. Readme is attached which
shows how to invoke gluster_test_main.

Fixes: #70


nishith-vihar:
Sorry about the PR message. I forgot to remove the template comments

nik-redhat:
Please update the commit-msg as per the template.

parsing/gluster_params_handler.py
```diff
@@ -0,0 +1,105 @@
...
+        throughout the class.
+        Args:
+            filepath (str): Path for config file
+        """
+        cls.config_hashmap = Parser.generate_config_hashmap(filepath)
+
+    @classmethod
+    def get_server_ip(cls, server_name: str) -> str:
+        """
+        Gives the server ip given the server name
```
nik-redhat, line 30: Since you are explaining what this function returns, this line is not required.
Update in other functions as well.

nik-redhat: Or, rather keep it but modify the return `docstring` a bit to be differentiatied.

nik-redhat: For example:
```
def get_all_bricks(mnode, volname):
    """Get list of all the bricks of the specified volume.

    Args:
        mnode (str): Node on which command has to be executed
        volname (str): Name of the volume

    Returns:
        list: List of all the bricks of the volume on Success.
        NoneType: None on failure.
    """
```

parsing/gluster_params_handler.py
```diff
@@ -0,0 +1,105 @@
...
+
+    @classmethod
+    def get_brick_root_list(cls, server_name: str) -> list:
+        """
+        Returns the list of brick root given the server name
+        Args:
+            server_name (str): name of the server as in config file.
+                         example: server_vm1
+        Returns:
+            ilist of brick root of the given server
```
nik-redhat, line 87: Typo: list

parsing/gluster_test_main.py
```diff
@@ -0,0 +1,37 @@
...
+    All the apis for the required functions will be called here
+    """
+    args = pars_args()
+
+    if args.config_file:
+        ParamsHandler.get_config_hashmap(args.config_file)
+
+
+if __name__ == '__main__':
+    main()
```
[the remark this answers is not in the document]
nishith-vihar: Yeah this is fine actually

nik-redhat:
Looks good, but let's wait for Srijan's review for the config file structure.

parsing/config.yml
```diff
@@ -0,0 +1,362 @@
...
+# Windows), super_user name(Admin/Administrator/* in case of windows or root/*
+# in case of linux).
+# This section has to be defined.
+clients_info:
+    *client_vm1: &client1
+    *client_vm2: &client2
+        super_user: 'root'
+    *client_vm3: &client3
+        super_user: 'Admin'
+        platform: 'windows'
```
srijan-sivakumar, line 72: I'm not sure we are supporting windows right out of the box. Maybe make it linux based distro.

srijan-sivakumar: This is wrt this test framework. Glusterfs can be mounted and used by Windows client. But for now, let's focus on nix based clients.

parsing/README.md
```diff
@@ -0,0 +1,10 @@
+# test-framework
+automation test framework for testing glusterfs
+
+Usage:
+
+Following command return the config hashmap after loading the config file.</br>
+-c option is required for specifying the config file.</br>
+The gluster_test_parser library parses the config file.
+
+> python3 gluster_test_main.py -c <<-path to config file->>
```
srijan-sivakumar, line 10: instead of `gluster_test_main.py`, maybe we can name it inline with the framework name and test_runner. So something like `redant_test_runner.py`

parsing/gluster_test_main.py
```diff
@@ -0,0 +1,37 @@
+"""
+This module is the gluster_test_main module which
+takes care of:
+1) Config file parsing (by gluster_test_parser).
+2) Tests-to-run list preparation (by test_list_builder).
+3) Invocation of the component which manages test execution (test_runner).
```
srijan-sivakumar, line 6: Too many levels of calling and execution. If we go by this then the flow would be, gluster_test_main -> test_runner -> runner_thread wherein the test runner would just get the list of all tests and it'll invoke the runner_thread. test_runner then just feels like an extra level being added in here. Parsing, Test list preparation and invocation of runner thread accordingly should be inside the test_runner.py.

srijan-sivakumar: In this manner, this will become our test runner. Why to add one extra level of abstraction ? Any particular reason behind it ?

srijan-sivakumar (changes requested):

parsing/gluster_test_main.py
srijan-sivakumar: One more point, if this is the script to be called at the very beginning, it should be at the base directory of the project an not inside the parsing.

nishith-vihar: > Too many levels of calling and execution. If we go by this then the flow would be, gluster_test_main -> test_runner -> runner_thread wherein the test runner would just get the list of all tests and it'll invoke the runner_thread. test_runner then just feels like an extra level being added in here. Parsing, Test list preparation and invocation of runner thread accordingly should be inside the test_runner.py.

Actually according to the design doc there are both test_runner and a test_main but for now i can make the main as test_runner

srijan-sivakumar: Cool, but what are your thoughts ? Do we actually see a need for another module being invoked in between which would just read the list and call runner_thread or should the test_main and test_runner be fused into one module to handle the same?

nishith-vihar: > Cool, but what are your thoughts ? Do we actually see a need for another module being invoked in between which would just read the list and call runner_thread or should the test_main and test_runner be fused into one module to handle the same?

At this point of time i am not sure about how multiple threads will be handled. If the test_runner only invokes threads then i guess there is no need for an extra layer but if it has to handle some other cases in case of multiple threads then maybe a seperate module for running threads would be fine.

srijan-sivakumar: Firstly, I think the nomenclature is still problematic. The runner_thread won't be executed as a thread. If we had to deal with threads, python would be the last place to create a solution into. The runner_thread would be invoked as a exec function from within the test_runner and then passed on the parameters. So basically a standard way of how parent and child process functions in linux.

nishith-vihar: Got that

parsing/config.yml
```diff
@@ -0,0 +1,362 @@
+log_file: /var/log/tests/gluster_tests.log
+
+# 'servers' is list of Hostnames/IP's of servers in the cluster.
+# This section has to be defined.
+servers:
+    - &server_vm1 server-vm1
```
srijan-sivakumar, line 6: maybe you can add examples for the configurations which need to be added in a comment. For now we might be able to understand as to what goes where but down the line, for people starting out, this config file will be a maze and difficult to tame without examples.

parsing/README.md
srijan-sivakumar: `redant_test_runner.py` ?

srijan-sivakumar (changes requested):
Better to add a readme supporting the config file and as to what all goes into the config file and what all information needs to be added by a user who's going to use this framework.

redant_test_runner.py
```diff
@@ -0,0 +1,37 @@
+"""
+This module is the gluster_test_main module which
+takes care of:
+1) Config file parsing (by gluster_test_parser).
+2) Tests-to-run list preparation (by test_list_builder).
+3) Invocation of the runner_thread which runs the testcase.
```
aujjwal-redhat, line 6: Invocation of test runner right? Or are we going to perform the test runner work also in this ?

nishith-vihar: As of now we are don't have a seperate component as test_runner and the gluster_test_main in design doc is renamed as redant_test_runner and this component will directly bring up threads using runner_thread.

aujjwal-redhat: Ack.

parsing/CONFIG_README.md
```diff
@@ -0,0 +1,37 @@
...
+
+<h3>1. servers</h3>
+The servers section consists of all server ip addresses which run the glusterd
+service and provide a storage environment to which the clients mount their 
+filesystem.
+A single server in the config file has to be specified with a reference name 
+followed by ip address in the following way:<br>
+`&server_vm1 0.0.0.0`<br>
+server_vm1 is a reference to its ip address which can be used to refer to its 
+ip address in the further config file or in the test cases. 1 should be replaced
```
srijan-sivakumar, line 15: Nit, instead of `1`, `one`

srijan-sivakumar (commented):
Would the server and client credentials be mentioned in this config file ? Or would that be a different config file..

nishith-vihar:
> Would the server and client credentials be mentioned in this config file ? Or would that be a different config file..

But Actually barak had said that the user has to insert only the ip. is there any idea of implementing a passwordless ssh or do we continue with credentials. I think even glusto does not take any credentials other than ip for ssh

srijan-sivakumar:
Hmm..ok. We can go by the passwordless ssh manner also, That would work. It's just that the user would have to set it up by themselves. It is but an one time activity. But I feel, we should be having the mechanism to accept password based connections too. Or we can make it a strict passwordless ssh only manner. @nishith-vihar , @neesingh-rh , @aujjwal-redhat , @nik-redhat  and @BarakSason  what are your thoughts on this ? Should we enforce passwordless ssh or keep it both passwordless ssh as well as authentication with password with every ssh.

nik-redhat:
I think the passwordless ssh approach would be better.
But, to have a flexible framework if we can accommodate both, would be great!

aujjwal-redhat:
> I think the passwordless ssh approach would be better.
> But, to have a flexible framework if we can accommodate both, would be great!

Yeah I would suggest the same.

nishith-vihar:
I have added the credentials for one server and client just for example for now and can be removed safely once passwordless ssh has been implemented

srijan-sivakumar (approved):

srijan-sivakumar:
LGTM. But we need to take this up and fine tune as we move ahead.

Issue #75, opened 2021-03-15
Create PARENT_TEST


aujjwal-redhat:
Create parent_test to be integrated with the rest of the components.

PR #76, opened 2021-03-15
Added parent_test

aujjwal-redhat:
Started working on parent_test
Created the basic functions and added the standard info.

Updates: #75

### DIFF
--- a/parent_test.py
+++ b/parent_test.py
@@ -1,0 +1,58 @@
+from abc import ABC
+
+
+class Parent_Test(ABC):
+
+    """
+    This class contains the standard info and methods which are needed by
+    most of the tests
+    """
+
+    _PASS: bool = True
+    _FAIL: bool = False
+
+    _TEST_RES: bool = _PASS
+
+    COMPONENT: str
+    TEST_NAME: str
+
+    def configure(self, component_arg: str, test_name_arg: str):
+        """
+        Initializes COMPONENT and TEST_NAME
+        """
+        self.COMPONENT = component_arg
+        self.TEST_NAME = test_name_arg
+
+    def core_init(self):
+        """
+        Establishes ssh connection
+        Since added in redant_mixin so
+        can call establish_connection directly
+        """
+
+        self.establish_connection()
+
+    def init(self):
+        """
+        Establishes connection by calling core_init
+        Creates volume
+        And runs the specific component in the
+        test case
+        """
+
+        self.core_init()
+        self.rlog(f"{self.TEST_NAME} from {self.COMPONENT} inits")
+
+    def run_test(self):
+        """
+        It will be overriden by Test cases
+        """
+        pass
+
+    def terminate(self):
+        """
+        Deletes the volume
+        Closes connection
+        """
+
+        self.rlog(f"{self.TEST_NAME} terminates")
--- a/parent_test.py
+++ b/parent_test.py
@@ -11,8 +11,6 @@ class ParentTest:
 
     TEST_RES: states the result of the test case
 
-    COMPONENT: component which is to be tested
-    TEST_NAME: name of the test to run
     """
 
     def __init__(self, passed: bool):
--- a/parent_test.py
+++ b/parent_test.py
@@ -1,5 +1,3 @@
-# from redant_libs.redant_mixin import RedantMixin as rm
-
 
 class ParentTest:
 
@@ -13,13 +11,14 @@ class ParentTest:
 
     """
 
-    def __init__(self, passed: bool):
+    def __init__(self, redant: object):
         """
         Creates volume
         And runs the specific component in the
         test case
         """
-        self.PASS = passed
+        self.PASS = True
+        self.redant = redant
         self.TEST_RES = self.PASS
 
     def run_test(self):
--- a/parent_test.py
+++ b/parent_test.py
@@ -16,12 +16,6 @@ class Parent_Test(ABC):
     COMPONENT: str
     TEST_NAME: str
 
-    def configure(self, component_arg: str, test_name_arg: str):
-        """
-        Initializes COMPONENT and TEST_NAME
-        """
-        self.COMPONENT = component_arg
-        self.TEST_NAME = test_name_arg
 
     def core_init(self):
         """
--- a/parent_test.py
+++ b/parent_test.py
@@ -1,39 +1,47 @@
-from abc import ABC
+from redant_libs.redant_mixin import RedantMixin as rm
 
 
-class Parent_Test(ABC):
+class Parent_Test:
 
     """
     This class contains the standard info and methods which are needed by
     most of the tests
+
+    PASS: states the pass value
+
+    TEST_RES: states the result of the test case
+
+    COMPONENT: component which is to be tested
+    TEST_NAME: name of the test to run
     """
 
-    _PASS: bool = True
-    _FAIL: bool = False
+    PASS: bool
 
-    _TEST_RES: bool = _PASS
+    TEST_RES: bool
 
     COMPONENT: str
     TEST_NAME: str
 
-    def init(self):
+    def __init__(self, passed: bool, component: str, test_name: str):
         """
         Creates volume
         And runs the specific component in the
         test case
         """
-        self.rlog(f"{self.TEST_NAME} from {self.COMPONENT} inits")
+        self.PASS = passed
+        self.TEST_RES = self.PASS
+        self.COMPONENT = component
+        self. TEST_NAME = test_name
+        rm.rlog(f"{self.TEST_NAME} from {self.COMPONENT} inits")
+
+    def init(self):
+        pass
 
     def run_test(self):
-        """
-        It will be overriden by Test cases
-        """
         pass
 
     def terminate(self):
         """
-        Deletes the volume
-        Closes connection
+        Closes connection for now.
         """
-
-        self.rlog(f"{self.TEST_NAME} terminates")
+        rm.rlog(f"{self.TEST_NAME} terminates")
--- a/parent_test.py
+++ b/parent_test.py
@@ -17,14 +17,6 @@ class Parent_Test(ABC):
     TEST_NAME: str
 
 
-    def core_init(self):
-        """
-        Establishes ssh connection
-        Since added in redant_mixin so
-        can call establish_connection directly
-        """
-
-        self.establish_connection()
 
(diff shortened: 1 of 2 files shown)


parent_test.py
```diff
@@ -0,0 +1,58 @@
...
+    TEST_NAME: str
+
+    def configure(self, component_arg: str, test_name_arg: str):
+        """
+        Initializes COMPONENT and TEST_NAME
+        """
+        self.COMPONENT = component_arg
+        self.TEST_NAME = test_name_arg
+
+    def core_init(self):
```
srijan-sivakumar, line 26: Not required.

srijan-sivakumar: This is taken care in runner_thread itself.

srijan-sivakumar (changes requested):

parent_test.py
aujjwal-redhat: Ok. Will remove it then.

parent_test.py
```diff
@@ -0,0 +1,42 @@
+from abc import ABC
```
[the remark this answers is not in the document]
aujjwal-redhat: abstract class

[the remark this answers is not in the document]
srijan-sivakumar: Also, do add a module doc, addressing as to what is happening in here.

parent_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class Parent_Test(ABC):
```
aujjwal-redhat, line 4: @srijan-sivakumar here I am extending abstract class properties.

srijan-sivakumar: Ack.

parent_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def run_test(self):
+        """
+        It will be overriden by Test cases
+        """
+        pass
+
+    def terminate(self):
+        """
+        Deletes the volume
```
srijan-sivakumar, line 35: let's not say volume as there are TCs which won;t even be creating volumes.

parent_test.py
```diff
@@ -0,0 +1,39 @@
...
+    _FAIL: bool = False
+
+    _TEST_RES: bool = _PASS
+
+    COMPONENT: str
+    TEST_NAME: str
+
+    def init(self):
+        """
+        Creates volume
```
srijan-sivakumar, line 21: Same as below. Volume creation is not in purview in every TC. Let's just say initialization before the TC func is run.

redant_libs/redant_mixin.py
```diff
@@ -6,10 +6,12 @@
 from redant_libs.support_libs.relog import Logger
 from redant_libs.gluster_libs.peer_ops import peer_ops
 from redant_libs.gluster_libs.volume_ops import VolumeOps
-
+from parent_test import Parent_Test
```
srijan-sivakumar, line 9: why does mixin need the parent class ? It is the test cases which would need the parent test to be inherited right?

srijan-sivakumar (changes requested):

redant_libs/redant_mixin.py
aujjwal-redhat: Yeah true

aujjwal-redhat: So should I import the logging module inside the parent class only ? I added it in redant_mixin to use the logging specifically.

srijan-sivakumar: We can safely remove the logging from the abstract class. If at all any user needs logging that the TC has been started that can be overridden inside the child class. Also, each TC will have it's own logging file so, there won't be any confusion as to when the logs pertaining to a certain TC start and end.

aujjwal-redhat: If we remove logging from abstract class then how will we show init() and terminate() function executed?

```js
self.rlog(f"{self.TEST_NAME} from {self.COMPONENT} inits")
```

srijan-sivakumar: technically, do we need to show them being executed ? Because the only thing we can do is invoke the init and terminate of the child class and call the init and child of it's super ( parent ), in this way we don't need logging to be added as part of the parent class

srijan-sivakumar: Also, what all operations do you think can generically be part of the init and terminate in the parent class ?

aujjwal-redhat: > technically, do we need to show them being executed ? Because the only thing we can do is invoke the init and terminate of the child class and call the init and child of it's super ( parent ), in this way we don't need logging to be added as part of the parent class

> Also, what all operations do you think can generically be part of the init and terminate in the parent class ?

init operation must consist of starting gluster in the server
since few operations like peer ops don't need volume to be created so even I feel volume creation should not be part of init.

But if gluster didn't start because of some issue then we will need logging ryt?

srijan-sivakumar: So, here's the thing. We will be running tests parallely in a cluster. So, by doing starting or stopping glusterfs inside init and terminate, we'd be disruption the whole env. This should be part of the TC. Secondly, the order in which the TC will be run will be helpful to not do certain checks again and again. Why do we have to re-create volumes again and again if the test is about something else altogether. Same goes for glusterd start and stop. Technically, they'd be part of a TC body and not init and terminate. In giving init and terminate so much functionality, we are repeating what past framework has been doing so, where do we save the compute and resource here ?

srijan-sivakumar: @nik-redhat what's your thought on this ?

aujjwal-redhat: Then why do we need the parent test class to be abstract ?

aujjwal-redhat: I think there is a confusion coz what we have received as input from the java demo in prev to prev meet says this:
```

	public void core_init() {
		// Init logger
		// Init ssh connection
	}

	public void init() {
		core_init();
		// Create connection
		// Create volume
		System.out.println(test_name + " from component " + component + " inits");
	}

	public void terminate() {
		// Delete volume
		// Close connection
		System.out.println(test_name + " terminates");
	}
}
```

aujjwal-redhat: Even rhe doc says this:
```
parent_test (5) - This component contains all common functionality and data which is required by all tests. 
This component inherits functionality from the components in the next sections.
Examples of shared functionality and data may be:
* Implementation of init and terminate methods which create and destroy the environment in which the test operates
* Variables to contain common data such as test name, whether this test can run parallely to other tests, test result, etc’
```

Let's discuss this in the meet.

parent_test.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self, passed: bool, component: str, test_name: str):
+        """
+        Creates volume
+        And runs the specific component in the
+        test case
+        """
+        self.PASS = passed
+        self.TEST_RES = self.PASS
+        self.COMPONENT = component
+        self. TEST_NAME = test_name
```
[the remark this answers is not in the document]
aujjwal-redhat: Yeah true

parent_test.py
```diff
@@ -0,0 +1,47 @@
...
+        rm.rlog(f"{self.TEST_NAME} from {self.COMPONENT} inits")
+
+    def init(self):
```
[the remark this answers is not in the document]
aujjwal-redhat: Yeah. Will remove it.

parent_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+    PASS: states the pass value
+
+    TEST_RES: states the result of the test case
+
+    COMPONENT: component which is to be tested
+    TEST_NAME: name of the test to run
+    """
+
+    PASS: bool
```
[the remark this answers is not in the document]
aujjwal-redhat: Yeah.

parent_test.py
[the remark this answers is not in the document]
aujjwal-redhat: Should I remove the test name and component completely as the thread_runner is also taking care?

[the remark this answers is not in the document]
srijan-sivakumar: Better!

parent_test.py
srijan-sivakumar: @aujjwal-redhat , I think you forgot the mixin object here :p

srijan-sivakumar (changes requested):
The name mixin object also is a part of the parent class. Do check the latest changes pushed by Nishith for the naming convention to be used for this object.

aujjwal-redhat:
> The name mixin object also is a part of the parent class. Do check the latest changes pushed by Nishith for the naming convention to be used for this object.

Oh ok.

srijan-sivakumar (approved):

srijan-sivakumar:
@aujjwal-redhat , just a nit..Could you modify the existing sample_tc.py and use this Parent class in the same ?

aujjwal-redhat:
> @aujjwal-redhat , just a nit..Could you modify the existing sample_tc.py and use this Parent class in the same ?

Yeah sure

aujjwal-redhat:
@srijan-sivakumar I have modified the sample_test.py file.
Called the sample_test.py file from another class; tested it with dummy variables.
```js
PASS = True
Redant = {'name': 'ayush'}
Test result = True
```
Got the above as the output

tests/sample_test.py
```diff
@@ -23,7 +25,7 @@ def __init__(self, redant: object):
                              Point of contact for the redant
                              framework.
         """
-        self.redant = redant
+        super().__init__(redant=redant)
```
srijan-sivakumar, line 28: is this required ? Won't the child class automatically invoke the parent constructor ?

srijan-sivakumar (changes requested):

tests/sample_test.py
aujjwal-redhat: so should we delete the __init__ of the test clas?

srijan-sivakumar: Yes...in this scenario, it won't be required.

aujjwal-redhat: I would suggest to remove it coz automatically the constructor is working the same in parent_Test. Moreover, if we aren't going to override anything else in the TestClass's init then it would be better to remove it.

aujjwal-redhat: >Yes...in this scenario, it won't be required.

Ok will not remove it completely then

srijan-sivakumar (approved):

Issue #80, opened 2021-03-16
Handling all exceptions in Rexe


srijan-sivakumar:
The current flow in rexe doesn't handle all exceptions and hence changing it to handle all the exceptions.

PR #81, opened 2021-03-16
Modified exception to handle all exceptions.

srijan-sivakumar:
A return code way doesn't work with Paramiko and
hence added generic exception.

Fixes: #80

### DIFF
--- a/redant_libs/support_libs/rexe.py
+++ b/redant_libs/support_libs/rexe.py
@@ -70,8 +70,8 @@ class Rexe:
                     username=self.host_user,
                     password=self.host_passwd)
                 self.rlog(f"SSH connection to {node} is successful.", 'D')
-            except paramiko.ssh_exception.AuthenticationException:
-                self.rlog("Authentication failure. Please check conf.", 'E')
+            except Exception as e:
+                self.rlog(f"Connection failure. Exception : {e}", 'E')
                 self.connect_flag = False
             self.node_dict[node] = node_ssh_client
 


nik-redhat (approved):

nik-redhat:
@srijan-sivakumar any idea what could be the reason for the import error in pylint?

srijan-sivakumar:
> @srijan-sivakumar any idea what could be the reason for the import error in pylint?

I assume we might not be following the module standards. individual files don't have any issue with pylint, but when we are running pylint on the while directory that's when we see this issue. I'm checking on how to handle it.

nishith-vihar:
> > @srijan-sivakumar any idea what could be the reason for the import error in pylint?
> 
> I assume we might not be following the module standards. individual files don't have any issue with pylint, but when we are running pylint on the while directory that's when we see this issue. I'm checking on how to handle it.

I think we need to add __init__.py in the directory so that it will be considered as a module to import.

nishith-vihar (approved):

srijan-sivakumar:
> > > @srijan-sivakumar any idea what could be the reason for the import error in pylint?
> > 
> > 
> > I assume we might not be following the module standards. individual files don't have any issue with pylint, but when we are running pylint on the while directory that's when we see this issue. I'm checking on how to handle it.
> 
> I think we need to add **init**.py in the directory so that it will be considered as a module to import.
Have you tried it ?

nishith-vihar:
> > > > @srijan-sivakumar any idea what could be the reason for the import error in pylint?
> > > 
> > > 
> > > I assume we might not be following the module standards. individual files don't have any issue with pylint, but when we are running pylint on the while directory that's when we see this issue. I'm checking on how to handle it.
> > 
> > 
> > I think we need to add **init**.py in the directory so that it will be considered as a module to import.
> 
> Have you tried it ?

Yeah. But doesn't seem to work

Issue #85, opened 2021-03-17
Test_List_Builder to be added


aujjwal-redhat:
Add test_list_builder to create a list of  tests-to-run to be used by the redant_test_runner component.

PR #86, opened 2021-03-17
Added test_list_builder

aujjwal-redhat:
Added test_list_builder.
Rn it creates a list of tests and returns the same

Updates: #85

### DIFF
--- a/test_list_builder.py
+++ b/test_list_builder.py
@@ -1,0 +1,49 @@
+"""
+
+This component is responsible for creating a list
+of tests-to-run to be used by the redant_test_runner component.
+Input for this component : the tests-to-run list which was parsed
+from the cli and the return: list of all the tests which are going
+to be executed in the current session.
+
+"""
+import os
+import glob
+
+
+class TestListBuilder:
+    """
+    This class contains create_test_list
+    which helps in creeting a list of
+    tests that has to be run
+    """
+
+    @classmethod
+    def create_test_list(cls, dir_path: list) -> list:
+        """
+        This function creates a list of test cases
+        that are going to be executed and returns
+        the same.
+        """
+        list_of_tests: list = []
+        tests_to_run: list = []
+
+        try:
+
+            for path in dir_path:
+                abs_dir_path: str = os.path.abspath(path)
+
+                if not abs_dir_path.endswith(".py"):
+                    list_of_files: list = glob.glob(f"{abs_dir_path}/*.py")
+
+                    list_of_tests.append(list_of_files)
+                else:
+                    list_of_tests.append([abs_dir_path])
+
+            for test in list_of_tests:
+                tests_to_run.extend(test)
+
+        except Exception as e:
+            print(e)
+
+        return tests_to_run
--- a/test_list_builder.py
+++ b/test_list_builder.py
@@ -25,7 +25,6 @@ class TestListBuilder:
         that are going to be executed and returns
         the same.
         """
-        list_of_tests: list = []
         tests_to_run: list = []
 
         try:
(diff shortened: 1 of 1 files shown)


test_list_builder.py
```diff
@@ -0,0 +1,49 @@
...
+
+class TestListBuilder:
+    """
+    This class contains create_test_list
+    which helps in creeting a list of
+    tests that has to be run
+    """
+
+    @classmethod
+    def create_test_list(cls, dir_path: list) -> list:
```
srijan-sivakumar, line 22: I'm assuming that the `dir_path` would be the tests directory so that it'll just pick the test files rather than lib files.

test_list_builder.py
```diff
@@ -0,0 +1,49 @@
...
+        """
+        This function creates a list of test cases
+        that are going to be executed and returns
+        the same.
+        """
+        list_of_tests: list = []
```
srijan-sivakumar, line 28: Why are we maintaining to different lists here ? Any design consideration ?

srijan-sivakumar (commented):
What all things do you thing comes under the TestListBuilder ? This looks like a start but we need to jot those functionalities so as to be clear as to how this module will evolve.

test_list_builder.py
aujjwal-redhat: Yeah it is the path for test directory.
Like I considered we will be passing something like:
dir_path = ['tests/function/afr', 'tests/function/bvt']

and it will return all the test cases for afr and bvt and if we pass even the specific file.py it will add that to the list.

test_list_builder.py
aujjwal-redhat: Yeah. I will extend it in a single list.

aujjwal-redhat:
> What all things do you thing comes under the TestListBuilder ? This looks like a start but we need to jot those functionalities so as to be clear as to how this module will evolve.

The test list builder should be just focussed for building the test cases list as it appears in the doc. But I think we can instead create a list of lists of test cases, where outer list should consist of two inner lists: 
1. sequential list: it consist of those test cases that need to run sequentially like test_add_brick, test_delete_brick. Coz brick need to be added then only it can be deleted.
2. parallel: which can run in parallel. Theses tests will be independent of one another.

That's just an analysis. Rest is how things go forward.

srijan-sivakumar:
> > What all things do you thing comes under the TestListBuilder ? This looks like a start but we need to jot those functionalities so as to be clear as to how this module will evolve.
> 
> The test list builder should be just focussed for building the test cases list as it appears in the doc. But I think we can instead create a list of lists of test cases, where outer list should consist of two inner lists:
> 
>     1. sequential list: it consist of those test cases that need to run sequentially like test_add_brick, test_delete_brick. Coz brick need to be added then only it can be deleted.
> 
>     2. parallel: which can run in parallel. Theses tests will be independent of one another.
> 
> 
> That's just an analysis. Rest is how things go forward.

So, the test list builder must be able to determine as to what TCs can be run sequentially and what can be run parallely..How do you propose we achieve this ?

test_list_builder.py
srijan-sivakumar: Have you tried this out ? Try creating a sample test_runner and invoking this inside that file and see if the desired result is achieved.

aujjwal-redhat: Yeah.

I created a mock_runner_file:
```js
from test_list_builder import TestListBuilder

if __name__ == "__main__":

    tests_to_run = TestListBuilder.create_test_list(['tests/afr','tests/bvt'])

    for test in tests_to_run:
        print(f"{test} is running\n")
        
```

This is the result I got:
```js
(redantenv) [root@serverc redant]# python3 mockt_runner.py 
/home/redant/tests/afr/test1.py is running

/home/redant/tests/afr/test2.py is running

/home/redant/tests/bvt/test1.py is running

/home/redant/tests/bvt/test2.py is running
```

aujjwal-redhat: @srijan-sivakumar lemme know if this is fine.

srijan-sivakumar: LGTM

srijan-sivakumar (approved):

Issue #58, opened 2021-03-12
Lint and Flake errors fixing


nishith-vihar:
All the files have to fixed for lint and flake errors. This issue can be referred for all the PRs fixing lint and flake.

srijan-sivakumar:
Same as #89 hence closing this.

PR #88, opened 2021-03-17
Created a pylintrc file

aujjwal-redhat:
Created a .pylintrc file using 
```js
pylint --generate-rcfile > .pylintrc
```
Now we can modify the same for the warnings that are not actually errors

The warnings that I removed by adding them in the .pylintrc :
```js
disable=E1101,
        E0401,
        R0801,
        R0903,
        C0103,
        W0703,
        R0902
```
Updates: #58

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -20,7 +20,7 @@ jobs:
           pip install flake8
           pip install pylint
           if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
-          touch __init__.py 
+
       - name: Lint with flake8
         run: |
           flake8 $(pwd) --count --select=E9,F63,F7,F82 --show-source --statistics
@@ -28,4 +28,5 @@ jobs:
       - name: Lint with Pylint
         run: |
           pylint $(pwd)
-          rm __init__.py
+          pylint $(pwd)/redant_libs/
+
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -29,4 +29,5 @@ jobs:
         run: |
           pylint $(pwd)
           pylint $(pwd)/redant_libs/
+          pylint $(pwd)/tests/
 
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,593 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Specify a score threshold to be exceeded before program exits with error.
+# fail-under=10.0
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+# limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+# load-plugins=
+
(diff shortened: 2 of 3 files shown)


.github/workflows/linting.yml
```diff
@@ -20,12 +20,13 @@ jobs:
...
+
       - name: Lint with flake8
         run: |
           flake8 $(pwd) --count --select=E9,F63,F7,F82 --show-source --statistics
           flake8 $(pwd) --count --max-complexity=10 --max-line-length=79 --statistics
       - name: Lint with Pylint
         run: |
           pylint $(pwd)
-          rm __init__.py
+          pylint $(pwd)/redant_libs/
```
aujjwal-redhat, line 31: This is temporary. We need to create packages. Then we won't need to provide absolute paths.

nik-redhat:
Could you also fix the errors being reported by the pylint in the workflow?

aujjwal-redhat:
> Could you also fix the errors being reported by the pylint in the checks?

Yeah sure

aujjwal-redhat:
Errors & Messages:

E1101:
'%s %r has no %r member%s'
https://pycodequ.al/docs/pylint-messages/e1101-no-member.html

In our code
E1101: Instance of 'VolumeOps' has no 'rlog' member (no-member)

E0401:
 Unable to import 'redant_libs.support_libs.rexe' (import-error)

R0801:
R0801: Similar lines in 5 files

R0903,

C0103,
C0103: Variable name "f" doesn't conform to snake_case naming style (invalid-name)
W0703:
W0703: Catching too general exception Exception (broad-except)

R0902
R0902: Too many instance attributes (12/7) (too-many-instance-attributes)

These were not errors actually. 
Few among these which I felt we can still have are:
W0703: Catching too general exception Exception (broad-except)
Coz we might need to have more specific exceptions rather than
```js
except Exception as e
```
and
C0103: Variable name "f" doesn't conform to snake_case naming style (invalid-name)
Coz it might help in making the code more consistent throughout the project.

.pylintrc
```diff
@@ -0,0 +1,594 @@
...
+        attribute-defined-outside-init,
+        missing-docstring,
+        E1101,
+        E0401,
+        R0801,
+        R0903,
+        C0103,
+        W0703,
+        R0902
+
```
aujjwal-redhat, line 145: Here are the warnings that are disabled.

nik-redhat:
Is the import error that was being reported earlier fixed, or disabled?

aujjwal-redhat:
> Is the import error that was being reported earlier fixed, or disabled?

It is disabled for now.
The issue is we need to first create packages or either add pythonpath.
Adding pytohn path will not be a good choice as compared to creating packages.
That's why added the following too:
https://github.com/srijan-sivakumar/redant/pull/88#discussion_r595923628

aujjwal-redhat:
To regenerate the specific errors:
1. Remove the specific errors or all of them from disable in .pylintrc 
2. Then run the command:
```js
pylint -j 4 --rcfile=.pylintrc redant_libs/
```

nik-redhat:
Okay, do we have an issue open to track this? So, that we don't forget to fix it.

aujjwal-redhat:
> Okay, do we have an issue open to track this? So, that we don't forget to fix it.

Yeah I will add one with all the required info.

nik-redhat:
@srijan-sivakumar does it look fine to you?

aujjwal-redhat:
Added the issue #89 
@nik-redhat

.github/workflows/linting.yml
srijan-sivakumar: What about the test cases ?

redant_libs/support_libs/rexe.py
```diff
@@ -83,16 +83,14 @@ def execute_command(self, node, cmd):
         if not self.connect_flag:
             ret_dict['Flag'] = False
             return ret_dict
-        stdin, stdout, stderr = self.node_dict[node].exec_command(cmd)
```
srijan-sivakumar, line 86: No. this is not the way! rather it should be `_, stdout, stderr = self.node_dict[node].exec(cmd)`

srijan-sivakumar (changes requested):

.github/workflows/linting.yml
aujjwal-redhat: Added now.

srijan-sivakumar (approved):

Issue #91, opened 2021-03-18
Adding gluster Ops library


nishith-vihar:
The gluster Ops library will mainly contain operations for starting and stopping gluster service on the server.

PR #92, opened 2021-03-18
Adding GlusterOps library

nishith-vihar:
Gluster Ops lib is added which has functionalities
for stating and stopping glusterd service.

Fixes: #91

### DIFF
--- a/redant_libs/gluster_libs/gluster_ops.py
+++ b/redant_libs/gluster_libs/gluster_ops.py
@@ -1,0 +1,50 @@
+"""
+This file contains one class - GlusterOps wich holds
+operations on the glusterd service on the server
+or the client.
+"""
+
+
+class GlusterOps:
+    """
+    GlusterOps class provides APIs to start and stop
+    the glusterd service on either the client or the sever.
+    """
+
+    def gluster_start(self, node: str):
+        """
+        Starts the glusterd service on the specified node.
+        Args:
+            node (str): The node on which the glusterd service
+                        is to be started.
+        """
+        cmd = "systemctl start glusterd"
+
+        self.rlog(f"Running {cmd} on {node}")
+
+        ret = self.execute_command(node=node, cmd=cmd)
+
+        if int(ret['error_code']) != 0:
+            self.rlog(ret['error_msg'], 'E')
+            raise Exception(ret['error_msg'])
+
+        self.rlog("Successfully ran {cmd} on {node}")
+
+    def gluster_stop(self, node: str):
+        """
+        Stops the glusterd service on the specified node.
+        Args:
+            node (str): The node on which the glusterd service
+                        is to be stopped.
+        """
+        cmd = "systemctl sto glusterd"
+
+        self.rlog(f"Running {cmd} on {node}")
+
+        ret = self.execute_command(node=node, cmd=cmd)
+        print(ret)
+        if int(ret['error_code']) != 0:
+            self.rlog(ret['error_msg'], 'E')
+            raise Exception(ret['error_msg'])
+
+        self.rlog("Successfully ran {cmd} on {node}")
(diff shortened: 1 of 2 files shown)


redant_libs/gluster_libs/gluster_ops.py
```diff
@@ -0,0 +1,50 @@
...
+        self.rlog("Successfully ran {cmd} on {node}")
+
+    def gluster_stop(self, node: str):
+        """
+        Stops the glusterd service on the specified node.
+        Args:
+            node (str): The node on which the glusterd service
+                        is to be stopped.
+        """
+        cmd = "systemctl sto glusterd"
```
nik-redhat, line 40: nit: stop

redant_libs/gluster_libs/gluster_ops.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+
+class GlusterOps:
+    """
+    GlusterOps class provides APIs to start and stop
+    the glusterd service on either the client or the sever.
+    """
+
+    def gluster_start(self, node: str):
```
nik-redhat, line 14: make it `glusterd_start`. Same for below function as well

redant_libs/redant_mixin.py
```diff
@@ -7,10 +7,11 @@
 from redant_libs.support_libs.io_ops import io_ops
 from redant_libs.gluster_libs.peer_ops import peer_ops
 from redant_libs.gluster_libs.volume_ops import VolumeOps
+from redant_libs.gluster_libs.gluster_ops import GlusterOps
```
nik-redhat, line 10: why this difference in naming style?
`gluster_ops` looks better, and the same format should be used for `volume_ops`.

nishith-vihar: The naming convention for a class should be in PascalCase like GlusterOps and not like gluster_ops. So io_ops and peer_ops should also be changed accordingly.

nik-redhat: Ohh right, forgot that part :smile: 
Then, do update their naming to PascalCase.

redant_libs/gluster_libs/gluster_ops.py
srijan-sivakumar: +1

nishith-vihar: Fixed

redant_libs/gluster_libs/gluster_ops.py
nishith-vihar: Fixed

redant_libs/gluster_libs/gluster_ops.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+
+
+class GlusterOps:
+    """
+    GlusterOps class provides APIs to start and stop
+    the glusterd service on either the client or the sever.
+    """
+
+    def glusterd_start(self, node: str):
```
srijan-sivakumar, line 14: Have you tried stopping and starting glusterd more than 5 times? Do try that operation. Also add the status function.

srijan-sivakumar (changes requested):

redant_libs/gluster_libs/gluster_ops.py
nishith-vihar: Do i return a boolean value for status: True for active and false for inactive?

nishith-vihar: I will make a seperate testcase for starting and stopping glusterd >5 times

nik-redhat: Actually, we need many other libraries as well, like sometimes glusterd doesn't start immediately it takes time, check status and all. I thought it can be enhanced in the next iteration so skipped that.
But, only the reset failure can be done, if it's needed now.

srijan-sivakumar: Cool. Then as we go.

srijan-sivakumar (approved):

Issue #106, opened 2021-03-22
remote_exec class variable in test case to be changed


nishith-vihar:
the remote_exec class variable in the test_case can be easily misinterpreted as remote executioner whereas in the actual sense it the reference of the mixin object passed to the test case be the runner_thread. So the variable has to be changed accordingly.

PR #107, opened 2021-03-22
remote_exec class variable changed to redant

nishith-vihar:
The remote_exec class variable has been changed to
redant which refers to the mixin object passed as reference
to the test case and forms a point of contact to the redant
framework from the test case.

Fixes: #106

### DIFF
--- a/tests/sample_test.py
+++ b/tests/sample_test.py
@@ -1,0 +1,39 @@
+"""
+This file contains the format of a test case.
+It contains one class - TestCase wich would
+hold the functions to be run in the test case.
+"""
+
+
+class TestClass:
+
+    """
+    The TestCase class contains the functionsto be
+    run in the test case. Every function should contain
+    the flow of the function(APIs which are called) in
+    the form of steps.
+    """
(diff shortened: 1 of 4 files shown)


tests/sample_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class TestClass:
+
+    """
+    The TestCase class contains the functionsto be
+    run in the test case. Every function should contain
+    the flow of the function(APIs which are called) in
+    the form of steps.
+    """
```
srijan-sivakumar, line 15: We already have a sample_tc. Maybe we do these changes in that TC?

srijan-sivakumar: It is of the name `test_sample_tc`

nishith-vihar: This is the same file. I changed the file name. It was test_sample_tc.py

srijan-sivakumar: Cool.

srijan-sivakumar (approved):

Issue #103, opened 2021-03-22
Enhancing redant_test_main


srijan-sivakumar:
The redant_test_main has the following functionalities under it,

- [x] Parsing of the config file
- [x] Invocation of the test list builder and forming the order of test execution
- [x] Using the above two data to invoke the test_runner.

Now to achieve this, following parameters have to be provided to the redant_test_main during invocation as command line arguments,
1. Config file path
2. Test directory ( which contains the test cases or tree of test cases within )
3. Log file directory ( within which the log files for each TC will be created. )

srijan-sivakumar:
Basic set of enhancements done. Focus is now to push for TCs and then see if things fail, raise issues for the test_main.

PR #108, opened 2021-03-22
Enhancing redant_test_main

srijan-sivakumar:
Code changes:
1. Added file accessibilty check for config file.
2. Test list builder invocation added inside the main.

Updates: #103

### DIFF
--- a/redant_test_main.py
+++ b/redant_test_main.py
@@ -7,14 +7,39 @@ This module takes care of:
 
 import argparse
 from parsing.redant_params_handler import ParamsHandler
+from test_list_builder import TestListBuilder
+
+
+def is_file_accessible(path: str, mode: str='r') -> bool:
+    """
+    To check if the given file is accessible or not
+    so as to prevent parsing failures.
+    Args:
+        path (str): The file path whose accessibility is
+                    to be checked.
+        mode (str): The file access mode which would needs
+                    to be validated for the given path.
+    Returns:
+        True: The file at the given path is accessible for
+              the specified mode.
+        False: The file is not accessible for the specified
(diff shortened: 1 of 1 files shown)


redant_test_main.py
```diff
@@ -10,7 +10,7 @@
 from test_list_builder import TestListBuilder
 
 
-def is_file_accessible(path: str, mode: str='r') -> bool:
+def is_file_accessible(path: str, mode: str = 'r') -> bool:
```
nishith-vihar, line 13: There is an issue mentioned by Barak in redant MOMs regarding the is_file_accessible function. I think it should first be discussed in the meeting.

srijan-sivakumar: The issue raised in MOM is about optimal manner but wrt pythonic way of checking if a file is accessible, the above function handles the case optimally. Do you have any point wit regards to the implementation of this function ?

nishith-vihar: No. Nothing else

nishith-vihar (approved):

redant_test_main.py
```diff
@@ -38,8 +63,15 @@ def main():
...
     args = pars_args()
 
-    if args.config_file:
+    if is_file_accessible(args.config_file):
         ParamsHandler.get_config_hashmap(args.config_file)
+    else:
+        print(f"The config file at {args.config_file} is not accessible.")
+        return -1
+
+    test_cases_list = TestListBuilder.create_test_list([args.test_dir])
```
nishith-vihar, line 72: Is args.test_dir a single string consisting of multiple test dirs?

PR #119, opened 2021-03-23
Modified Parent_Test and runner_thread

aujjwal-redhat:
Parent_test:
============
Modified init and configure

Runner_thread:
==============
Removed the mixin call using RE_MIXIN and modified TC_OBJ as per the changes in Parent_test

Updates: #75

### DIFF
--- a/tests/parent_test.py
+++ b/tests/parent_test.py
@@ -1,3 +1,4 @@
+from redant_libs.redant_mixin import RedantMixin
 
 class ParentTest:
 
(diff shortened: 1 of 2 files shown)


tests/parent_test.py
```diff
@@ -1,25 +1,35 @@
...
 
 class ParentTest:
 
     """
     This class contains the standard info and methods which are needed by
     most of the tests
 
-    PASS: states the pass value
-
+    PASS: states the pass value i.e True
```
srijan-sivakumar: nit. We have only only TEST_RES

srijan-sivakumar (changes requested):
Just remove the detail about the pass and fail form the class doc.

tests/parent_test.py
srijan-sivakumar: Now..you pushed a commit before I could comment :p

srijan-sivakumar (approved):

tests/parent_test.py
aujjwal-redhat: XD sorry

Issue #118, opened 2021-03-23
Adding Functionality in Params handler


nishith-vihar:
APIs which return servers and clients info which consists of ip,hostname,user,passwd have to be added

PR #120, opened 2021-03-23
Added function to return nodes info

nishith-vihar:
The function return a dictionary with nodes info consisting
of hostname,ip,user and passwd.

Fixes: #118

### DIFF
--- a/parsing/redant_params_handler.py
+++ b/parsing/redant_params_handler.py
@@ -77,6 +77,76 @@ class ParamsHandler:
         return client_ip_list
 
     @classmethod
+    def get_nodes_info(cls) -> dict:
+        """
+        Returns a dictionary consisting of server info
+        Returns:
+            dict: dictionary consisting of server info
+        format of dictionary:
+        {
+            servers: [
+                        {
+                            "hostname" : server_vm1
+                            "ip" : 10.4.28.93
+                            "user" : root
+                            "passwd" : redhat
+                        },
+                        {
+                            "hostname" : server_vm2
+                            "ip" : 23.43.12.87
+                            "user" : root
+                            "passwd" : redhat
+                        }
+            ]
+            clients: [
+                        {
+                            "hostname" : client_vm1
+                            "ip" : 10.3.28.92
+                            "user" : root
+                            "passwd" : redhat
+                        },
+                        {
+                            "hostname" : client_vm2
+                            "ip" : 15.12.43.98
+                            "user" : root
+                            "passwd" : redhat
+                        }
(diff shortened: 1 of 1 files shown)


parsing/redant_params_handler.py
```diff
@@ -76,6 +76,76 @@ def get_client_ip_list(cls) -> list:
...
+                            passwd : passwd
+                        },
+                        {
+                            hostname : hostname
+                            ip : ip
+                            user : user
+                            passwd : passwd
+                        }
+           ]
+        }
```
nik-redhat, line 115: nit: Can rephrase it to example format and instead provide some dummy values to make it more clear.
Like,
```
{
                            hostname : node1
                            ip : 10.70.1.121
                            user : root
                            passwd : password
                        },
```

parsing/redant_params_handler.py
```diff
@@ -76,6 +76,76 @@ def get_client_ip_list(cls) -> list:
...
+    def get_nodes_info(cls) -> dict:
+        """
+        Returns a dictionary consisting of server info
+        Returns:
+            dict: dictionary consisting of server info
+        format of dictionary:
+        {
+            servers: [
+                        {
+                            hostname : hostname
```
srijan-sivakumar, line 89: "hostname" . These are json values. Keep the keys are strings. So it'd be better to reflect the same in example.

srijan-sivakumar (changes requested):
nit: Change in the example.

srijan-sivakumar:
> How come we have lint failures in other files, if it is being checked in all the PRs?

@nik-redhat , currently we can ignore the lint failures. Let's finish the integration of the components and then we can take a breath and fix these linting issues.

parsing/redant_params_handler.py
```diff
@@ -76,6 +76,76 @@ def get_client_ip_list(cls) -> list:
...
+    def get_nodes_info(cls) -> dict:
+        """
+        Returns a dictionary consisting of server info
+        Returns:
+            dict: dictionary consisting of server info
+        format of dictionary:
+        {
+            servers: [
+                        {
+                            "hostname" : server_vm1
```
srijan-sivakumar, line 89: @nishith-vihar , even the values are strings..not just the keys..

srijan-sivakumar: Anyway, I'm merging it. This change can be made in a separate patch. This fix is essential to the integration of the main, test_runner, and the runner_thread

srijan-sivakumar (approved):

Issue #121, opened 2021-03-23
Logging to be re-structurized


srijan-sivakumar:
Logging is important for debugging the carcass of the TCs which fail. Hence it would be annoying to go through a data dump containing logs for all the TC to check why it failed. The better way would be to mimic the tests director structure and create a per volume type run per TC log file. So, no more going over a big file and using grep. Just navigate to that particular log file belonging to a TC.

PR #126, opened 2021-03-24
Logging re-structurized

srijan-sivakumar:
1. Logging class cleaned up.
2. Logging of TCs in their respective log files.
3. The logging dir structure resembles the TC
structure.
4. Added flag for specific test run but not
handled.
5. Test run based on multiple volume types
handled.
6. Output of redant_test_runner improved.

Fixes: #121


srijan-sivakumar:
Do ignore linting for now..We'll address that separately. Functional level changes are what is being done here.

Issue #127, opened 2021-03-24
Making changes to test cases


nishith-vihar:
The IPs are taken from the config file and and are accessed through Parent Test. The test case should be running successfully along with other components.

srijan-sivakumar:
We can close this as we have handled it from the TC side.

PR #129, opened 2021-03-24
glusterd_start_stop tc working successfully

nishith-vihar:
the tc accesses server from parent test and
works successfully along with other components

Updates: #127

### DIFF
--- a/tests/functional/glusterd/test_glusterd_start_stop.py
+++ b/tests/functional/glusterd/test_glusterd_start_stop.py
@@ -13,16 +13,17 @@ class TestCase(ParentTest):
     for glusterd service operations
     """
 
-    def gluster_start_stop_test(self):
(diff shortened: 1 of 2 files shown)


parsing/test_config.yml
```diff
@@ -0,0 +1,23 @@
+log_file: /var/log/tests/gluster_tests.log
+
+servers:
+    - &server_vm1 10.70.43.63 #Example- &server_vm1 0.0.0.0
```
srijan-sivakumar, line 4: The test config is not to be part of the PR :p

srijan-sivakumar (changes requested):
LGTM. Just remove the config.yml. Because it is not generic. Users using the TC framework should mention their own IPs and details.

srijan-sivakumar (approved):

PR #134, opened 2021-03-25
Peer probe and volume test case working

nishith-vihar:
Peer probe and volume test cases are working
successfully integrating with all the components.

Updates: #127

### DIFF
--- a/tests/functional/glusterd/test_peer_probe_detach.py
+++ b/tests/functional/glusterd/test_peer_probe_detach.py
@@ -36,4 +36,5 @@ class TestCase(ParentTest):
             self.redant.glusterd_stop(server1)
 
         except Exception as e:
+            self.TEST_RES = False
             print(f"Test is failed:{e}")
--- a/tests/functional/glusterd/test_peer_probe_detach.py
+++ b/tests/functional/glusterd/test_peer_probe_detach.py
@@ -1,7 +1,7 @@
 """
 This component has a test-case for peers addition and deletion.
 """
-#nonDisruptive;dist,rep,arb,disp,dist-rep,dist-arb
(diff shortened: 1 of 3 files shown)


tests/functional/glusterd/test_volume_create_delete.py
```diff
@@ -13,7 +13,7 @@ class TestCase(ParentTest):
     for volume creation and deletion
     """
 
-    def volume_create_delete_test(self):
+    def run_test(self):
```
nik-redhat, line 16: Why the change in function name?
Moreover, it is same as the previous test case

nishith-vihar: the runner_thread called the run_test function which is in parent test. So here we are overriding the run_test function from parent test so that the runner thread call the run_test function in the test case.

nishith-vihar: As of now all the test cases will have run_test  as the function name

tests/functional/glusterd/test_peer_probe_detach.py
```diff
@@ -22,16 +22,18 @@ def test_peer_probe_detach(self):
...
+            self.redant.pool_list(server1)
 
-            self.redant.peer_detach("192.168.122.220", "192.168.122.161")
+            self.redant.peer_detach(server1, server2)
 
-            self.redant.glusterd_stop("192.168.122.220")
+            self.redant.glusterd_stop(server1)
 
         except Exception as e:
             print(f"Test is failed:{e}")
```
neesingh-rh, line 39: I think we can now handle assignment of TEST_RES with Fail also in exception as parent class is now inherited.

srijan-sivakumar (approved):

neesingh-rh (approved):

Issue #135, opened 2021-03-25
Adding more functionality in ops libraries.


neesingh-rh:
Add functionalities to ops libraries along with the test cases to verify them.
1. In io_ops:- rmdir for removing directories and file_exist function to check if a file exists or not
2. In peer_ops:-  nodes_from_pool_list and get_pool_list for retrieving the nodes in the pool

srijan-sivakumar:
Hi @neesingh-rh , can you modify the issue description to define as to what functionality is being added.

neesingh-rh:
> Hi @neesingh-rh , can you modify the issue description to define as to what functionality is being added.

Okay.

srijan-sivakumar:
Closing this issue as we have to modularly track issues rather than a simple issue saying `add more functionality`.

srijan-sivakumar:
@nishith-vihar @neesingh-rh @aujjwal-redhat . Take this as a consideration. Vague issue names and descriptions are not a good indicator and would lead to confusion to the reviewers. Start creating issues specific to tasks which are directed towards some functional addition and one can understand the feature right from the issue being fixed/added and the description before even moving to the code.

PR #141, opened 2021-03-29
Functionalities are added in ops lib

neesingh-rh:
io_ops now contains only execute_io_cmd function

Updates: #135 
Signed-off-by: Neeraj Pratap Singh neesingh@redhat.com


srijan-sivakumar (changes requested):
Please rebase and send out the patch. Also, can you add some supporting TCs to verify these functionalities ?

neesingh-rh:
> Please rebase and send out the patch. Also, can you add some supporting TCs to verify these functionalities ?

Yeah..I will add TCs .

srijan-sivakumar (commented):
@neesingh-rh , just rebase and check if there are any conflicts with the test_comparison.py

neesingh-rh:
> @neesingh-rh , just rebase and check if there are any conflicts with the test_comparison.py

There are no conflicts.

neesingh-rh:
I have added the new test case with the modifications to handle non-volume TCs.I did run the TC with modifications, it is running.

support/ops/gluster_ops/peer_ops.py
```diff
@@ -150,3 +150,51 @@ def pool_list(self, node: str):
...
+
+    def get_pool_list(self, node: str):
+        """
+        Parse the output of 'gluster pool list' command.
+
+        Args:
+            node (str): Node on which command has to be executed.
+
+        Returns:
+            list : list of dicts on success.
```
nishith-vihar, line 182: You can give an example format of the list of dicts returned.

tests/example/test_comparison.py
```diff
@@ -0,0 +1,64 @@
+"""
```
nishith-vihar, line 1: Why is the test_comparison file added again example?

tests/functional/glusterd/test_comparison.py
```diff
@@ -2,7 +2,7 @@
 This file contains a test-case which tests glusterd
 starting and stopping of glusterd service.
 """
-#disruptive;dist
+# disruptive;dist
```
nishith-vihar, line 5: Space should not be added in the volume types comment as it will be parsed by comment parser.

tests/example/test_file_creation.py
```diff
@@ -0,0 +1,67 @@
+"""
```
nishith-vihar, line 1: Why is the same file added in both example and glusterd?

srijan-sivakumar (approved):

Issue #150, opened 2021-03-31
Enhancing the config file


nishith-vihar:
mount information in the config file has to be removed and servers allocated to the volumes have to be removed

srijan-sivakumar:
Also moving it to /config from /core/parsing

nishith-vihar:
host name field has to be added in the servers info

nishith-vihar:
Config file is simplified to clients info,servers info and volume types. Parsing module should be changed accordingly.

PR #151, opened 2021-03-31
Enhancing the config file

nishith-vihar:
The mount info is removed and the servers section is removed
from the volume info part.

Fixes: #150


BarakSason:
I think lines 144 to 146 can be removed as well.
In addition, please move the config file from /core/congif to /config.

BarakSason (changes requested):
I think lines 144 to 146 can be removed as well.
In addition, please move the config file from /core/parsing to /config.

srijan-sivakumar (changes requested):
L69-L70 can be removed I think. The comment says `this is to define what volume types and mount protocols will be run in this current test run.`

srijan-sivakumar (approved):

Issue #156, opened 2021-04-05
Test case for comparing performance


nishith-vihar:
Test case which is equivalent to the java version has to be created which helps in comparison between the java and python versions

nishith-vihar:
Unmount volume functionality is added in VolumeOps library

PR #157, opened 2021-04-05
Comparision test_case added and unmount volume in VolumeOps

nishith-vihar:
test case for comparision of performance is addded and
unmount volume funtion is added in volumeOps

Fixes: #156


srijan-sivakumar (approved):
LGTM

Issue #159, opened 2021-04-07
Creating cluster funtion to be added


nishith-vihar:
A function which creates a cluster given the cluster servers has to be added in the peer ops library.

PR #160, opened 2021-04-07
Cluster creation funtionality added

nishith-vihar:
Function to create a cluster given a list of servers
is added in the peer ops library

Fixes: #159


srijan-sivakumar (changes requested):
@nishith-vihar rebase this and also add a TC to support this addition.

srijan-sivakumar (approved):

Issue #161, opened 2021-04-08
Handle glusterd start stop burst


srijan-sivakumar:
A function needs to be added which handles the burst failure of glusterd.

This is the command of the form : `systemctl reset-failed glusterd`

Issue #163, opened 2021-04-08
Add Functionality to gluster ops


srijan-sivakumar:
Following functionality to be added to the glusterd ops.

1. wait_for_glusterd_to_start
2. get_gluster_version

PR #164, opened 2021-04-08
Gluster-ops enhancement

srijan-sivakumar:
Following code changes have been handled:
1. Addition of reset-failed feature to
handle burst limit issue in glusterd.
2. Gluster ops functions enhanced
to handle multi server operations.
3. Two new functions added
    3.1 wait_for_glusterd_to_start
    3.2 get_gluster_version

Fixes: #161, #163

### DIFF
--- a/support/ops/gluster_ops/gluster_ops.py
+++ b/support/ops/gluster_ops/gluster_ops.py
@@ -3,7 +3,7 @@ This file contains one class - GlusterOps wich holds
 operations on the glusterd service on the server
 or the client.
 """
-
+from time import sleep
 
 class GlusterOps:
     """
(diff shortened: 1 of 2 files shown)


support/ops/gluster_ops/gluster_ops.py
```diff
@@ -3,53 +3,133 @@
...
+from time import sleep
 
 class GlusterOps:
     """
     GlusterOps class provides APIs to start and stop
     the glusterd service on either the client or the sever.
     """
 
-    def glusterd_start(self, node: str):
+    def start_glusterd(self, node, enable_retry: bool=True):
```
nishith-vihar, line 14: If we are randomizing the nodes I think the node should default to None

aujjwal-redhat:
LGTM. Just the issue with the failing check related to python build. Other than that everything looks fine.

srijan-sivakumar:
> LGTM. Just the issue with the failing check related to python build. Other than that everything looks fine.

Yeah..added the `self` keyword for the same in a particular function call.

support/ops/gluster_ops/gluster_ops.py
srijan-sivakumar: @nishith-vihar , Starting of glusterd cannot be randomized :p doesn't come into the category of commands which can be run from anywhere. For example, to start a volume, we can run it from any of the nodes in a cluster but to start glusterd in a node, we have to run the start_glusterd from that node only.

nishith-vihar: Got it

nishith-vihar:
LGTM

nishith-vihar (approved):

Issue #162, opened 2021-04-08
Handle non volume TCs


srijan-sivakumar:
It is not compulsory for TCs to have a volume type or set of volume types to be run or specifically.

Maybe the TC has to handle multiple TCs in the logic itself. In this scenario, the flag has to be NA. Which corresponds to don't assume the volume or rather don't give the flag.

PR #165, opened 2021-04-08
Handle non volume TCs.

srijan-sivakumar:
Now one, user can even push non volume TCs, i.e.
TC won't require to be linked to any particular
volume type. This leads to two scenarios,
1. A TC which doesn't even create volumes.
2. A TC which might go one to deal with
volumes of multiple times at a time.

Fixes: #162


nishith-vihar (approved):
LGTM

aujjwal-redhat (approved):
LGTM

Issue #167, opened 2021-04-08
Logger not logging module and line number


srijan-sivakumar:
The idea of having a generic logger function has backfired. As Tony Stark would say, "Not a great plan :/". Hence to revert back to old format of pythonic logging.

Sometimes it is better to stick to a logging methodology followed by the industry as there might be some reasoning as to why they do ( we sort of learnt the why in this experience. ).

srijan-sivakumar:
The logger function will be reverted back to logger.info, logger.debug and logger.error.

srijan-sivakumar:
For example : 

```
[2021-04-08 16:13:54,940] DEBUG [relog - 48:rlog] - {'msg': OrderedDict([('opRet', '0'), ('opErrno', '0'), ('opErrstr', None), ('output', 'success')]), 'Flag': True, 'node': 'server-vm2', 'cmd': 'gluster --xml peer detach server-vm3 --mode=script', 'error_code': 0}
[2021-04-08 16:13:54,941] INFO [relog - 48:rlog] - Successfully ran gluster --xml peer detach server-vm3 --mode=script on server-vm2
[2021-04-08 16:13:54,941] INFO [relog - 48:rlog] - Running gluster --xml peer detach server-vm1 --mode=script on node server-vm2
[2021-04-08 16:13:55,184] DEBUG [relog - 48:rlog] - {'msg': OrderedDict([('opRet', '0'), ('opErrno', '0'), ('opErrstr', None), ('output', 'success')]), 'Flag': True, 'node': 'server-vm2', 'cmd': 'gluster --xml peer detach server-vm1 --mode=script', 'error_code': 0}
[2021-04-08 16:13:55,184] INFO [relog - 48:rlog] - Successfully ran gluster --xml peer detach server-vm1 --mode=script on server-vm2
[2021-04-08 22:30:30,389] DEBUG [relog - 48:rlog] - establish connection
[2021-04-08 22:30:30,524] DEBUG [relog - 48:rlog] - SSH connection to server-vm1 is successful.
[2021-04-08 22:30:30,633] DEBUG [relog - 48:rlog] - SSH connection to server-vm2 is successful.
[2021-04-08 22:30:30,768] DEBUG [relog - 48:rlog] - SSH connection to server-vm3 is successful.
[2021-04-08 22:30:30,769] INFO [relog - 48:rlog] - Running gluster --xml peer probe server-vm1 on node server-vm2
[...]

PR #168, opened 2021-04-09
Logger doesn't reflect proper module and line number

srijan-sivakumar:
This seems to be related to the fact that the logger
takes in the function and line number values from
where it is being executed and hence we cannot
delegate and generalize the logger functions by
calling them from a common location.

This might be something which works in languages
which go through a good level of compilation steps
but the same is not done in python. Hence reverting back
to the pythonic way of logging.

Fixes: #167


nishith-vihar (approved):
LGTM

Issue #179, opened 2021-04-11
Parent test to be Abstract parent test


srijan-sivakumar:
It makes sense to convert the existing Parent test into the abstract test. Along with this change, following changes will be undertaken,
1. The try-except logic which earlier used to be part of the TC will now be handled inside a module ( `parent_run_test` )in the parent test which in turn will invoke the abstract test case ( `run_test` ) which will be overridden by the user.
2. The second functionality which is deemed to be added is including a hashmap ( dictionary ) containing relationship between volume type and the details required for creating a volume. This detail will be provided by the param handler. Now a user can either user the default values in the config file for volume related operations or they can override these details in a constructor inside the TC.

srijan-sivakumar:
With the recent discussions, following changes are to be incorporated,

- [ ] Cluster creation if volume type is not generic, followed by the creation of volume and it being started acc to config parameters if user doesn't override the default parameters.
- [ ]  Terminate function to handle the cleanup of the volume, cluster and the paramiko connection.

srijan-sivakumar:
Closing this Issue. Will open respective issues for further feature additions.

PR #180, opened 2021-04-11
Parent test now an abstract parent test

srijan-sivakumar:
The Parent test has been converted into
an abstract class with certain methods
being abstract.

Now, the following changes have been incorporated
in this patch:
1. Exception handling now part of parent test class
and hence the test cases need not concern themselves
of the try except logic.
2. The test cases have been modified to reflect this latest
change - wrt Pt. 1.
3. A test case removed.
4. Traceback has been added into the parent test functionality.
This will be added to log whenever an exception fails a TC.

Updates: #179


aujjwal-redhat (approved):
LGTM.

tests/functional/glusterd/test_peer_probe_detach.py
```diff
@@ -12,28 +13,14 @@ class TestCase(ParentTest):
...
-            server_list = [server1,server2]
-            self.redant.create_cluster(server_list)
-
-        except Exception as e:
-            self.TEST_RES = False
-            print(f"Test is failed:{e}")
+        redant.create_cluster(self.server_list)
+        redant.pool_list(server1)
+        redant.peer_detach(server1, server2)
+        redant.peer_detach(server2, server2)
```
nik-redhat, line 26: The build error for these line seems genuine :thinking: @srijan-sivakumar

srijan-sivakumar: Aaah..Yes..The details about server1 and server2 have to be taken up from the server_list. Will send out a commit.

nik-redhat (approved):
lgtm

Issue #182, opened 2021-04-12
Modify the peer ops


aujjwal-redhat:
* Remove the non-matching docstrings. Functions are returning ret which is not a tuple.
* Added the function type as per the standards.

Issue #172, opened 2021-04-09
Standardizing the Code


aujjwal-redhat:
* Ops:

1. docstring doesn't match the actual return:
Ex: peer_ops
```js
def peer_status(self, node: str):
        """
        Checks the status of the peers
        Args:
            node (str): Node on which command has to be executed.

        Returns:
            tuple: Tuple containing three elements (ret, out, err).
                The first element 'ret' is of type 'int' and
                is the return value
                of command execution.

                The second element 'out' is of type 'str'
                and is the stdout value
                of the command execution.

                The third element 'err' is of type 'str'
                and is the stderr value
                of the command execution.
        """

        cmd = 'gluster --xml peer status'

        self.logger.info(f"Running {cmd} on node {node}")

        ret = self.execute_command(node, cmd)

        if ret['error_code'] != 0:
            self.logger.error(ret['msg']['opErrstr'])
            raise Exception(ret['msg']['opErrstr'])

        self.logger.info(f"Successfully ran {cmd} on {node}")
        return ret
```
Here we are returning just ret which is a dictionary but the docstring is stating it as tuple.

2. In tests we are not using the ret value:
Test: peer_probe_detach
```js
def run_test(self):
        """
        In this testcase:
        1) glusterd service is started
        2) peer probe of a server
        3) list the storage pool
        4) peer detach
        5) glusterd is stopped
        """
        try:
            server1 = self.server_list[0]
            server2 = self.server_list[1]

            self.redant.start_glusterd(server1)

            self.redant.peer_probe(server2, server1)

            self.redant.pool_list(server1)

            self.redant.peer_detach(server1, server2)

            self.redant.stop_glusterd(server1)

        except Exception as e:
            self.TEST_RES = False
            print(f"Test is failed:{e}")
```
If we are not using the ret value then returning that might not be needed.

* Class names
In some classes SnakeCase is used while in some CamelCase. This is the initial stage so keeping the name formats standard will be better.

* Certain ops are not returning ret or anything.
```js
def volume_mount(self, server: str, volname: str,
                     path: str, node: str=None):
        """
        Mounts the gluster volume to the client's filesystem.
        Args:
            node (str): The client node in the cluster where volume
                        mount is to be run
            server (str): Hostname or IP address
            volname (str): Name of volume to be mounted
            path (str): The path of the mount directory(mount point)
        """

        cmd = f"mount -t glusterfs {server}:/{volname} {path}"

[...]

PR #183, opened 2021-04-12
Modified the peer ops file

aujjwal-redhat:
Added the return types in function.
Removed the non-matching docstrings.

Fixes: #182
Updates: #172

### DIFF
--- a/support/ops/gluster_ops/peer_ops.py
+++ b/support/ops/gluster_ops/peer_ops.py
@@ -26,6 +26,7 @@ class peer_ops:
                 - error_msg: error message
                 - error_code: error code returned
                 - cmd : command that got executed
+                - node : node on which the command got executed
         """
 
         cmd = f'gluster --xml peer probe {server}'
@@ -58,6 +59,8 @@ class peer_ops:
                 - error_msg: error message
                 - error_code: error code returned
                 - cmd : command that got executed
+                - node : node on which the command got executed
+
         """
 
         if force:
@@ -87,6 +90,8 @@ class peer_ops:
                 - error_msg: error message
                 - error_code: error code returned
                 - cmd : command that got executed
+                - node : node on which the command got executed
+
         """
 
         cmd = 'gluster --xml peer status'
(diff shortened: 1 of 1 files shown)


support/ops/gluster_ops/peer_ops.py
```diff
@@ -44,7 +31,7 @@ def peer_probe(self, server: str, node: str):
             raise Exception(ret['msg']['opErrstr'])
 
         self.logger.info(f"Successfully ran {cmd} on {node}")
-        return ret
```
srijan-sivakumar, line 47: Any specific reason as to why the return value is being removed?
What if the TC demands that the response of peer probe is to be checked if it remains the same after multiple peer probe and detach commands. There are certain glusto TCs which perform this type of operation.

support/ops/gluster_ops/peer_ops.py
```diff
@@ -83,27 +57,13 @@ def peer_detach(self, node: str, server: str, force: bool = False):
...
```
srijan-sivakumar, line 86: Same as above.

support/ops/gluster_ops/peer_ops.py
```diff
@@ -117,9 +77,9 @@ def peer_status(self, node: str):
...
```
srijan-sivakumar, line 120: Same as above.

srijan-sivakumar (changes requested):

support/ops/gluster_ops/peer_ops.py
aujjwal-redhat: Oh ok. Will add ret then.

support/ops/gluster_ops/peer_ops.py
aujjwal-redhat: Ack

neesingh-rh (approved):

support/ops/gluster_ops/peer_ops.py
aujjwal-redhat: @srijan-sivakumar is it fine now?

srijan-sivakumar (approved):
LGTM

Issue #193, opened 2021-04-15
Passing the whole config hashmap to parent test


nishith-vihar:
Instead of just passing the nodes info from the params handler the whole config hashmap has to be passed to the parent test as we also need volume types information to be passed to the parent test.

PR #194, opened 2021-04-15
Passing the config hashmap to the parent test.

nishith-vihar:
The config hashmap is passed from the params handler to the
parent test.

Fixes: #193

### DIFF
--- a/core/parsing/params_handler.py
+++ b/core/parsing/params_handler.py
@@ -13,7 +13,7 @@ class ParamsHandler:
     """
 
     @classmethod
-    def get_config_hashmap(cls, filepath: str):
+    def set_config_hashmap(cls, filepath: str):
         """
         Gets the configuration hashmap generated from the
         api in Parser class and sets the reuired class variable.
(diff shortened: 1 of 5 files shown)


tests/parent_test.py
```diff
@@ -13,13 +13,17 @@ class ParentTest(metaclass=abc.ABCMeta):
...
+                 log_path: str, log_level: str = 'I'):
         """
         Creates volume
         And runs the specific component in the
         test case
         """
+        
+        server_details = config_hashmap['servers_info']
+        client_details = config_hashmap['clients_info']
+        volume_types_info = config_hashmap['volume_types']
```
srijan-sivakumar, line 26: Can this be converted to a class variable ? As this might be used inside a TC also for invoking `volume_create` function.

srijan-sivakumar (changes requested):

tests/parent_test.py
nishith-vihar: Ack

srijan-sivakumar:
@neesingh-rh @aujjwal-redhat do add your review comments..

core/parsing/params_handler.py
```diff
@@ -46,49 +46,78 @@ def get_client_ip_list(cls) -> list:
...
-           }
-        }
         """
-
-        nodes_info = {}
-        s_info = cls.config_hashmap["servers_info"]
-        c_info = cls.config_hashmap["clients_info"]
-        
-        servers = list(s_info.keys())
+        config_hashmap = cls.config_hashmap
```
aujjwal-redhat, line 118: Do we really need to store the config_hashmap and return. Can't we just return it?

srijan-sivakumar: Adding to the above point, how many functions are we actively using in the framework from the ones we have defined in the params_handler.py ?

nishith-vihar: Yeah. Actually we are using only 1 function. That is the get_config_hashmap. I think all the other functions are reduntant but actually the design decision was that only relevant information be passed from the config hashmap after parsing by calling the functions in the params handler but with this change the params handler becomes insignificant

nishith-vihar: > Do we really need to store the config_hashmap and return. Can't we just return it?

No Its not required. That's just for readability. I can directly return if required

aujjwal-redhat: Ack

srijan-sivakumar: Couple of points,
1. If other functionalities are not being used anywhere ( I don't see a utility in near future ), maybe we can do a cleanup in this PR itself.
2. We can pass the value directly rather than storing it in a variable. It won't make much of a difference as such, except ofcourse a variable being created and then destroyed which points to that hashmap inside the function stack. But in terms of simplicity, let's remove the assignment and directly pass the value. The tool in use was created for making the loves of devs easier. So, let's follow the pythonic way.

srijan-sivakumar (approved):
LGTM

aujjwal-redhat (approved):
LGTM

Issue #198, opened 2021-04-15
Creating a readme for writing test cases


aujjwal-redhat:
Create a readme that can guide anyone to  integrate with the ops and write their own tests.

PR #199, opened 2021-04-15
Added procedure to write a test case

aujjwal-redhat:
Added the steps to write a test case from scratch

Fixes: #198

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,18 +1,18 @@
-# Test Cases.
+# Test Case
 
 1. [About](#about)
 2. [How to write a test case](#how-to-write-a-test-case)
 
 ## About
 The TCs are divided into
-1. Functional
-2. Performance
+1. Functional : These tests are basically to test the working of the functionalities.
+2. Performance : Performance tests deals with the performance i.e. *how fast is this working* :zap:
 
 Inside these categories we have sub-category depending on the gluster
 component being tested in the given TC.
 
 ### Points of consideration while writing TC:
-1. The naming of the TCs should be prefixed with test_
+1. The naming of the TCs should be prefixed with **test_**
 2. The type of the TC ( disruptive or non-disruptive ) and the
 types of volumes for which it'll be run has to be part of a TC. On referring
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -37,3 +41,43 @@ format
...
+This file contains a test-case which tests glusterd
+starting and stopping of glusterd service.
+
+```
+
+3. Add the test type(disruptive or non-disruptive) and volume.
+```js
+Disruptive: disruptive
+Non-disruptive: nonDisruptive
+```
```
neesingh-rh, line 62: We can give example for mentioning volume type along with this eg-disruptive;dist
Rest looks fine to me

aujjwal-redhat: Ack

aujjwal-redhat: Actually volume table is specified in the table above already so I will put it's reference here.

tests/README.md
```diff
@@ -37,3 +41,43 @@ format
...
+```js
+from tests.parent_test import ParentTest
+```
+
+5. Create a class that extends the `ParentTest`.
+```js
+class TestCase(ParentTest):
+```
+
+6. Make a function `run_test`. This function basically overrides the run test function from the [Parent Test](https://github.com/srijan-sivakumar/redant/blob/main/tests/parent_test.py)
```
srijan-sivakumar, line 75: How about we explain when to add __init__, and terminate also ? Because there might be cases wherein the user would have to override the existing functions ( Though we would have to invoke the parent test's constructor and terminate ourselves inside the TC ).

srijan-sivakumar (changes requested):
I'd say we write it an explanatory manner. Imagine yourself to be a user who doesn't know this framework and wants to write the TC. The steps are always good but maybe we can add a little bit of flair to the writing so that it isn't dry.

Maybe treat the readme like a blog..I think that might be a good approach here.

tests/README.md
aujjwal-redhat: Ack

srijan-sivakumar (approved):
:) LGTM

Issue #204, opened 2021-04-16
Fixing cluster creation and cluster deletion functions


srijan-sivakumar:
Proposed solution : [Cluster Doc](https://docs.google.com/document/d/1TlKnTqeE6jpE4kIpoWT60wWvwCYD1Trf9EE6uo_ZUQU/edit?usp=sharing)

PR #205, opened 2021-04-16
Fixing cluster creation and cluster deletion fn

srijan-sivakumar:
Fixes: #204

### DIFF
--- a/support/ops/gluster_ops/peer_ops.py
+++ b/support/ops/gluster_ops/peer_ops.py
@@ -3,7 +3,7 @@ This file contains one class - PeerOps which
 holds APIS related to peers which will be called
 from the test case.
 """
-
+import socket
 
 class PeerOps:
     """
@@ -42,7 +42,7 @@ class PeerOps:
 
         return ret
 
-    def peer_detach(self, node: str, server: str, force: bool = False):
+    def peer_detach(self, server: str, node: str, force: bool = False):
         """
         Detach the specified server.
         Args:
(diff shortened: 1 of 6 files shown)


srijan-sivakumar:
one set of change done is that the way how the cluster is created is modified. Instead of the way mentioned and proposed in the doc, I'm destroying the clusters and then re-creating it. This seems to be the simplified and straightforward way rather than taking a round-about approach and handling multiple failure cases in the documented approach.

support/ops/gluster_ops/peer_ops.py
```diff
@@ -155,6 +155,25 @@ def nodes_from_pool_list(self, node: str) -> list:
...
+        the try-except block.
+        If anyone were to come across a better syntax, do change this
+        monstrosity.
+        """
+        try:
+            if 'hostname' in pool_list_data.keys():
+                nodes.append(pool_list_data['hostname'])
+                return nodes
+        except:
+            pass
```
aujjwal-redhat, line 176: Should we just `pass` or add some message as well?

support/ops/gluster_ops/peer_ops.py
```diff
@@ -184,49 +203,87 @@ def get_pool_list(self, node: str) -> list:
...
+                node_list.remove(value)
+                node_list.append(ip_val)
+        return node_list
+
+    def create_cluster(self, node_list: list) -> bool:
+        """
+        Creates a cluster out of given set of nodes irrespective
+        of their existing cluster configurations.
+        Args:
+            node_list (list): All nodes which are to be part of the cluster.
```
aujjwal-redhat, line 248: NIT: Return details in docstring missing.

tests/parent_test.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, mname: str, config_hashmap: dict, volume_type: str,
         self.client_list = list(client_details.keys())
 
         self.redant.start_glusterd()
+        self.redant.create_cluster(self.server_list)
```
aujjwal-redhat, line 39: I guess the function referring to is `_add_orphan_nodes_to_cluster` . It was replaced in the above lines.

support/ops/gluster_ops/peer_ops.py
srijan-sivakumar: Ack. Will add that.

tests/parent_test.py
srijan-sivakumar: Yes, the function hasn't been removed. Will do a cleanup.

support/ops/gluster_ops/peer_ops.py
srijan-sivakumar: I think the comment detailed the reasoning behind the addition of an except clause. No need to log anything here.

aujjwal-redhat: Ack

aujjwal-redhat (approved):
LGTM

Issue #207, opened 2021-04-20
Adding a mechanism to exclude specific tests


srijan-sivakumar:
A flag is to be added so that test_list builder doesn't go ahead and add it in the test runs. This is useful for tests not considered suitable to be run because of a BUG or due to some other valid reason.

srijan-sivakumar:
@nishith-vihar add a doc link explaining the design decision taken here. This is an important feature from the user perspective while being run by the user.

BarakSason:
Based on the discussion on the approach, I believe the name of this issue should be changed to "Adding a mechanism to exclude tests"

nishith-vihar:
@BarakSason Ack

PR #211, opened 2021-04-21
Flag for skipping the test is added

nishith-vihar:
when 'skip-test' comment is added in the test case then
that particular test case is skipped and is not executed

Fixes: #207

### DIFF
--- a/config/CONFIG_README.md
+++ b/config/CONFIG_README.md
@@ -33,3 +33,14 @@ ip:<br>
 'volume_types' defines different volume types that we can create in
 gluster and minimum number of servers are assigned to each of the
 volume types to run the tests. This section is not defined by the user.
+
+<h3>4. excluded_tests</h3>
+'excluded_tests' is a list of tests given by the user which are excluded during
+the test run. Each element of the list consists of the complete path of the test
+case with respect to the redant directory.
+
+Example format of excluded_tests list:<br>
+
+excluded_tests:<br>
+    &nbsp;&nbsp;&nbsp;&nbsp; - tests/functional/glusterd/test1<br>
+    &nbsp;&nbsp;&nbsp;&nbsp; - tests/functional/glusterd/test2<br>
--- a/config/CONFIG_README.md
+++ b/config/CONFIG_README.md
@@ -39,6 +39,9 @@ volume types to run the tests. This section is not defined by the user.
 the test run. Each element of the list consists of the complete path of the test
 case with respect to the redant directory.
 
+Specific design consideration and more information on the feature is given in
+the doc link [here](https://docs.google.com/document/d/1T0Fu20WzH7jACE9wTVT_SEyfrcXKQUmEExpLKru6Ys0/edit?usp=sharing)
+
 Example format of excluded_tests list:<br>
 
 excluded_tests:<br>
--- a/config/config.yml
+++ b/config/config.yml
@@ -49,3 +49,8 @@ volume_types:
         replica_count: 2
         arbiter_count: 1
         transport: tcp
+        
+#excluded_tests - Tests which are excluded during the test run.
+excluded_tests:
+    - tests/functional/glusterd/test1
+    - tests/functional/glusterd/test2
--- a/core/parsing/params_handler.py
+++ b/core/parsing/params_handler.py
@@ -157,3 +157,12 @@ class ParamsHandler:
         for server in self.server_config:
             brick_roots[server] = self.server_config[server]['brick_root'][0]
         return brick_roots
+
+    def get_excluded_tests(self):
+        """
+        Gets a list of exluded tests from the config file.
+        Returns:
+            list : list of exluded tests
+        """
+
+        return self.config_hashmap["excluded_tests"]
--- a/core/redant_main.py
+++ b/core/redant_main.py
@@ -18,7 +18,6 @@ from environ import environ
 from common.relog import Logger
 
 
-
 def pars_args():
     """
     Function to handle command line parsing for the redant.
--- a/core/redant_main.py
+++ b/core/redant_main.py
@@ -73,7 +73,7 @@ def main():
         return
 
     # Building the test list and obtaining the TC details.
-    excluded_tests = config_hashmap["excluded_tests"]
+    excluded_tests = param_obj.get_excluded_tests()
     test_cases_tuple = TestListBuilder.create_test_dict(args.test_dir,
                                                         excluded_tests,
                                                         args.spec_test)
@@ -94,7 +94,7 @@ def main():
     env_obj.setup_env()
 
     # invoke the test_runner.
-    TestRunner.init(test_cases_dict, param_obj, args.log_dir,
+    TestRunner.init(test_cases_dict, param_obj, excluded_tests, args.log_dir,
                     args.log_level, args.concur_count)
     result_queue = TestRunner.run_tests()
 
--- a/core/redant_main.py
+++ b/core/redant_main.py
@@ -18,6 +18,7 @@ from environ import environ
 from common.relog import Logger
 
 
+
 def pars_args():
     """
     Function to handle command line parsing for the redant.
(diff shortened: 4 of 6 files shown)


BarakSason:
Can you please clarify under what circumstances we'd want to add a 'skip-test' comment to the test?

nishith-vihar:
I think we can use it in cases where we would want to skip 1 or 2 test cases or some test cases related to a particular component due to some bug which may not be corrected immediately and we want to test all other test cases.

nik-redhat:
Similar to the use of `flaky` in our `.t` tests.

BarakSason:
> Similar to the use of `flaky` in our `.t` tests.

That's exactly why I'm not of this - If there's a problem, this problem needs to be fixed, not ignored and "hidden" someplace no one will find, or care about.

> I think we can use it in cases where we would want to skip 1 or 2 test cases or some test cases related to a particular component due to some bug which may not be corrected immediately and we want to test all other test cases.

This takes things from another view.
If a given test was working fine up until now, and due to some recent change is now failing, what's the point of just ignoring this failure? The whole point is to prevent regression.
If this test was failing from the start (due to an actual bug in the product), this test shouldn't be merged until the product itself is fixed.

IMO, the only case when tests needs to be skipped is if the config provided by the user cannot satisfy this particular tests requirements (e,g, the test MUST HAVE 10 servers to operate, but the environment only contains 5).

nik-redhat:
> That's exactly why I'm not of this - If there's a problem, this problem needs to be fixed, not ignored and "hidden" someplace no one will find, or care about.
> 
> This takes things from another view.
> If a given test was working fine up until now, and due to some recent change (in the product's code) is now failing, what's the point of just ignoring this failure? The whole point is to prevent regression.

The thing is sometimes it's not easy to fix few failures and that might cause some delay, and if due to a particular test that might not hold much importance (or even if it does), should the whole pipeline be blocked? IMO I don't think that's valid.

> If this test was failing from the start (due to an actual bug in the product), this test shouldn't be merged until the product itself is fixed.

That I think is a different scenario.

> 
> IMO, the only case when tests needs to be skipped is if the config provided by the user cannot satisfy this particular tests requirements (e.g. the test MUST HAVE 10 servers to operate, but the environment only contains 5).

This is a good case, but not the only possible reason for this feature AFAIT.

srijan-sivakumar:
> > Similar to the use of `flaky` in our `.t` tests.
> 
> That's exactly why I'm not of this - If there's a problem, this problem needs to be fixed, not ignored and "hidden" someplace no one will find, or care about.
> 

Ack. But its not the case most of the times. Maybe the TC has to be modified or it is still being debugged. Or maybe there's an open bug on it being worked on. In those scenario's, unless and until it is fixed, I don't feel it might add to value to run it only to see the failure.

> > I think we can use it in cases where we would want to skip 1 or 2 test cases or some test cases related to a particular component due to some bug which may not be corrected immediately and we want to test all other test cases.
> 
> This takes things from another view.
> If a given test was working fine up until now, and due to some recent change (in the product's code) is now failing, what's the point of just ignoring this failure? The whole point is to prevent regression.
> If this test was failing from the start (due to an actual bug in the product), this test shouldn't be merged until the product itself is fixed.
> 
> IMO, the only case when tests needs to be skipped is if the config provided by the user cannot satisfy this particular tests requirements (e.g. the test MUST HAVE 10 servers to operate, but the environment only contains 5).

Let's take a middle ground here. How about we allow skipping of tests but then show the tests which have been skipped ? That might help us keep a tab on test cases with issues that have to be solved.

BarakSason:
The normal workflow (ideally) should be that once a test was merged, it should never be marked to be skipped.
If a test was merged and suddenly fails it can be due to one of the following:

1. The test had some inherent bug was was not detected initially (e.g. some race case) and was detected in larger testing
2. A code change in the products suddenly fails the test - in this case we have actual regression

The ideal way to handle both of these is to fix what's broken.
However, since the situation is not always ideal due to priorities, I can accept, that in certain cases, excluding a test might be necessary, however I disagree with the implementation presented here.
With this implementation, certain TCs will be marked to skipped and will just be forgotten (we see this very well in Gluster).
I suggest to handle this in a different way - Add an "exclude" section to the config file and add there what tests needs to be skipped - This way the information is always in front of the developer's eyes and things have a much lesser change of being forgotten.
Moreover, from a design and OOP perspective, it's not the test's responsibility to tell the framework to ignore it, it should be the other way around - The framework should "know" which tests to exclude.

> Let's take a middle ground here. How about we allow skipping of tests but then show the tests which have been skipped ? That might help us keep a tab on test cases with issues that have to be solved.

Lets differentiate between "skipped" and "excluded":

- Skipped = Was not able to execute due to lacking environment
- Excluded = Was not executed as the user marked this test to be excluded

srijan-sivakumar:
> The normal workflow (ideally) should be that once a test was merged, it should never be marked to be skipped.
> If a test was merged and suddenly fails it can be due to one of the following:
> 
>     1. The test had some inherent bug was was not detected initially (e.g. some race case) and was detected in larger testing
> 
>     2. A code change in the products suddenly fails the test - in this case we have actual regression
> 
> 
> The ideal way to handle both of these is to fix what's broken.
> However, since the situation is not always ideal due to priorities, I can accept, that in certain cases, excluding a test might be necessary, however I disagree with the implementation presented here.
> With this implementation, certain TCs will be marked to skipped and will just be forgotten (we see this very well in Gluster).
> I suggest to handle this in a different way - Add an "exclude" section to the config file and add there what tests needs to be skipped - This way the information is always in front of the developer's eyes and things have a much lesser change of being forgotten.

I feel this is similar to adding skip inside the test case. In the CI nobody would be going in to check the config most of the times. I would rather focus on the result presented after a test run which shows how many tests are skipped. This metrics is what can help in putting the onus on the development team to fix the issues.

> Moreover, from a design and OOP perspective, it's not the test's responsibility to tell the framework to ignore it, it should be the other way around - The framework should "know" which tests to exclude.
> 

I beg to differ in this case. The framework can only judge TCs to be skipped if the framework cannot provide the requisite environment for the TC to run. But if the failure is happening due to the underlying logic of the platform which in our case is glusterfs, then I don't think it is the framework which is responsible, the onus is on the test case.

> > Let's take a middle ground here. How about we allow skipping of tests but then show the tests which have been skipped ? That might help us keep a tab on test cases with issues that have to be solved.
> 
> Lets differentiate between "skipped" and "excluded":
> 
>     * Skipped = Was not able to execute due to lacking environment
> 
>     * Excluded = Was not executed as the user marked this test to be excluded

We can add the above classification but then again I do feel strongly that 
1. Skipped can be handled by framework given the fact that framework has the picture of the environment not the TCs and the decision to skip a test can be done on the go.
2. The Exclusion is not dynamic. But then again, I don't think adding the details in config would make us quicker look into the issue. But its rather the end result which we always take into consideration and the comparison iteratively will push for changes.

BarakSason:
> I feel this is similar to adding skip inside the test case. In the CI nobody would be going in to check the config most of the times. I would rather focus on the result presented after a test run which shows how many tests are skipped. This metrics is what can help in putting the onus on the development team to fix the issues.
> 

In that regard it will absolutely not matter whether a flag will exist in the TC, or the TC will be excluded through the config file - if this test was supposed to run and it's marked, it should appear as "excluded" (not skipped) in the results.

> I beg to differ in this case. The framework can only judge TCs to be skipped if the framework cannot provide the requisite environment for the TC to run. But if the failure is happening due to the underlying logic of the platform which in our case is glusterfs, then I don't think it is the framework which is responsible, the onus is on the test case.
> 

Sorry, not agreeing here.
TC is a set of operations, not more not less. Judgment is not TC responsibility.
If a TC needs to be excluded due to product issue, no change should be done in the TC, it should be marked externally. Why should the TC code (changing a flag is changing code) change?

> We can add the above classification but then again I do feel strongly that
> 
>     1. Skipped can be handled by framework given the fact that framework has the picture of the environment not the TCs and the decision to skip a test can be done on the go.
> 
>     2. The Exclusion is not dynamic. But then again, I don't think adding the details in config would make us quicker look into the issue. But its rather the end result which we always take into consideration and the comparison iteratively will push for changes.

1 - My intention was that skipping tests will be determined at runtime, this should be a framework operation, was that not clear?
2 - IMO the config file approach is preferable for the reasons I've mentioned.

srijan-sivakumar (changes requested):
@nishith-vihar rebase this on the current main and see what changes have been done wrt params handler.

Issue #221, opened 2021-04-23
Gluster ops complying AbstractOps


aujjwal-redhat:
Gluster Ops still does not extend the abstract ops functionalities.

srijan-sivakumar:
The decision taken is to take a middle road here. We cannot be too stringent and worship the frameworks and the software methodology but at the same time we cannot throw everything out of the window just because we want to push changes fast. We will be using abstract ops ( seriously..somebody rename it, it is too misleading ) but at the same time, only at places wherein we don't have to compromise on the functionality in the ops function. Hence converting this into a discussion.

PR #222, opened 2021-04-26
Gluster ops modified to use abstract ops functionalities

aujjwal-redhat:
Replaced the redundant code in gluster ops with the abstract ops code.
Also the dispatcher was not able to recognise None type objects in case no node is passed. So added the `type(None)` in rexe in the dispatcher to handle the none case as well.
Fixes: #221

### DIFF
--- a/common/ops/abstract_ops.py
+++ b/common/ops/abstract_ops.py
@@ -24,7 +24,6 @@ class AbstractOps:
                         If the node is None then the rexe chooses the
                         node randomly and executes the command on it.
         """
-        self.logger.info(f"Running {cmd} on {node}")
 
         ret = self.execute_command(cmd, node)
 
@@ -36,9 +35,14 @@ class AbstractOps:
                 self.logger.error(ret['msg']['opErrstr'])
                 raise Exception(ret['msg']['opErrstr'])
 
+<<<<<<< HEAD
+=======
+        self.logger.info(f"Successfully ran {cmd} on {ret['node']}")
+
+>>>>>>> 16b28c6... Gluster ops modified to use abstract ops functionalities
         return ret
         
-    def execute_abstract_op_multinode(self, cmd : str, node : str=None):
+    def execute_abstract_op_multinode(self, cmd : str, node : list=None):
         """
         Calls the function in the remote executioner to execute
         commands on the nodes. Logging is also performed along
@@ -49,14 +53,14 @@ class AbstractOps:
                         If the node is None then the rexe chooses the
                         node randomly and executes the command on it.
         """
-        self.logger.info(f"Running {cmd} on {node}")
-
         ret = self.execute_command_multinode(cmd, node)
+        for each_ret in ret:
+            if each_ret['error_code'] != 0:
+                self.logger.error(each_ret['msg']['opErrstr'])
+                raise Exception(each_ret['msg']['opErrstr'])
 
-        if int(ret['msg']['opRet']) != 0:
-            self.logger.error(ret['msg']['opErrstr'])
-            raise Exception(ret['msg']['opErrstr'])
-
+            self.logger.info(f"Successfully ran {cmd} on {each_ret['node']}")
+            
         return ret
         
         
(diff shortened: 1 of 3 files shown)


common/ops/gluster_ops/gluster_ops.py
```diff
@@ -35,17 +35,10 @@ def start_glusterd(self, node=None, enable_retry: bool = True):
...
-        if node is None:
-            ret = self.execute_command_multinode(cmd)
-            self.logger.info(f"Running {cmd} on all nodes.")
-        else:
-            ret = self.execute_command_multinode(cmd, node)
-            self.logger.info(f"Running {cmd} on {node}")
-
+        ret = self.execute_abstract_op_multinode(cmd, node)
+    
         for result_val in ret:
```
nishith-vihar, line 40: I think this wont be required as ret will not be returned if error code is not zero instead exception will be raised from abstract ops

nishith-vihar: from line 40

common/ops/abstract_ops.py
```diff
@@ -49,14 +50,14 @@ def execute_abstract_op_multinode(self, cmd : str, node : str=None):
                         If the node is None then the rexe chooses the
                         node randomly and executes the command on it.
         """
-        self.logger.info(f"Running {cmd} on {node}")
-
         ret = self.execute_command_multinode(cmd, node)
+        for each_ret in ret:
+            if each_ret['error_code'] != 0:
+                self.logger.error(each_ret['msg']['opErrstr'])
+                raise Exception(each_ret['msg']['opErrstr'])
```
aujjwal-redhat, line 57: @nishith-vihar are you referring to this?

nishith-vihar: Yeah

aujjwal-redhat: Oh ok. Then should we also remove the part from execute_abstract_op():
```bash
if(ret['error_code']!=0):
            self.logger.error(ret['error_msg'])
            raise Exception(ret['error_msg'])    
        elif isinstance(ret['msg'],OrderedDict) or isinstance(ret['msg'],dict):
            if int(ret['msg']['opRet']) != 0:
                self.logger.error(ret['msg']['opErrstr'])
                raise Exception(ret['msg']['opErrstr'])
```

nishith-vihar: you do not need to remove anything from abstract ops. The abstract ops will raise the exception. We should remove from line 40 to 43 in gluster ops

nishith-vihar: because when exception is raised nothing is returned

aujjwal-redhat: Oh ok. I misunderstood the part above :) Will remove the part from gluster ops.

aujjwal-redhat:
@nishith-vihar But the `enable_retry` will still need the comm_fail check to restart glusterd.

nishith-vihar:
In that case either the abstract op should not raise an exception or the glusterd start function must not call the abstract function and directly call the execute_command_multinode function. This has to be discussed.

aujjwal-redhat:
> In that case either the abstract op should not raise an exception or the glusterd start function must not call the abstract function and directly call the execute_command_multinode function. This has to be discussed.

Yeah true.

common/ops/abstract_ops.py
nishith-vihar, line 57: I think this part should follow the pattern in the execute_abstract_op_node

aujjwal-redhat:
@nishith-vihar @srijan-sivakumar this pr won't be necessary is what I feel since as Nishith also mentioned that a few functions need ret value to work upon in the next steps. In that case gluster ops is fine as it is. Do lemme know your views. :)

srijan-sivakumar:
I think we are good to go ahead without this patch. As we have already seen, abstract ops save a lot of duplicate and to be honest frustrating code which we have to rinse and repeat everytime, but as all good features, we need to see where it fits. Certain functions cannot be put into a mold to fit the abstract ops requirements, therein we need to provide special treatment. Now I don't see a lot of functions going that path, but some will. The sane decision here would be to take a middle road, using abstract ops wherever possible ( and trying to keep functions modular ) and in very serious cases, just leave the function be.

srijan-sivakumar:
So, I'm closing this PR and do add comments if any,

Issue #209, opened 2021-04-20
Param handler usage to be based on object rather than class methods.


srijan-sivakumar:
The current issue with paramshandler is that the gc clears out the class variable but this is required for further operations across the framework.

Issue #186, opened 2021-04-12
Volume creation from parent TC based on config.


srijan-sivakumar:
The current TC flow doesn't give much automated feature for handling a TC. Technically a user shouldn't have to decide for how a volume is to be created but rather what type of volume and how many servers to be used for it.

The proposal is to handle it in the manner mentioned in the doc : [Volume Ops](https://docs.google.com/document/d/1K9Fuy8uP7LjUfZkajwzPVFBroZiW1uX5BF9c1pbVKvM/edit?usp=sharing)

srijan-sivakumar:
- [ ] Volume name to be taken from the testcase name and the volume type. Hence if the test case is `test_sample.py`, and of the type of volume for which it is being executed is `dist` then the name of the volume which'll be created will be `test_sample_dist`.
- [ ] Getting the server and brick details from the config file and hence the user doesn't have to pass the details on their own. Also, passing the volume options dictionary to the volume_create rather than having a kwars arrangement. This is two prong, a) Kwargs might consume more time and has more changes of exceptions and errors in runtime if not handled properly. b) The volume options being passed will be the one read from the config file and if the user doesn't override it with their own custom options then the default config will be used for creating the volume.
- [ ] We also need to track the volumes being created for a given TC. Hence a dictionary test_vol_dict will be passed on as an argument to the volume_create function which will add the volume details once it is successfully created. This same dictionary will come handy in deciding various other operations such as brick add, delete, replace, mount volume, unmount volume, volume cleanup/delete.
- [ ] A separate function called volume cleanup which will go through this said test_vol_list to to clean the test cluster. This is useful if we want the framework to handle the vol unmount, snap unmount,  snap deletion, vol stop, vol delete, brick cleanup etc in one single function.  This function will be available alongside the volume_stop and volume_delete.
- [ ] After the above operations are being handled properly introduce the dynamic selection of bricks. The current approach will name the bricks as per the volume and server name combination with a count. But if there are more than one brick root dirs then it'll be required to balance the brick dir creation across these given set of brick roots.
- [ ] It is not at all important to provide where to run many a gluster specific commands if they're part of a cluster. Hence the node to run the command can be left to the framework also.

nik-redhat:
> * [ ]  Volume name to be taken from the testcase name and the volume type. Hence if the test case is `test_sample.py`, and of the type of volume for which it is being executed is `dist` then the name of the volume which'll be created will be `test_sample_dist`.
[...]

PR #224, opened 2021-04-26
Volume creation in parent test and param handler changes

srijan-sivakumar:
Multiple code changes:
1. Volume creation method modified to handle replica,
distributed and distributed-replica volume with servers
having single brick root.
2. Param handler methods converted from class methods to
regular methods so that gc doesn't do a cleanup after
object goes out of scope.

Fixes: #209
Updates: #186

### DIFF
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -68,30 +68,19 @@ class VolumeOps(AbstractOps):
 
         return ret
 
-    def volume_create(self, volname: str, bricks_list: list,
-                      node: str = None, force: bool = False,
-                      **kwargs):
+    def volume_create(self, volname: str, node: str, conf_hash: dict,
+                      server_list: list, brick_root: list,
+                      force: bool = False):
         """
         Create the gluster volume with specified configuration
         Args:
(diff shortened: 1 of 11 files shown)


common/ops/gluster_ops/volume_ops.py
```diff
@@ -68,30 +68,19 @@ def volume_unmount(self, path: str, node: str = None):
...
-                    - arbiter_count : (int)|None
-                    - stripe_count : (int)|None
-                    - disperse_count : (int)|None
-                    - disperse_data_count : (int)|None
-                    - redundancy_count : (int)|None
-                    - transport_type : tcp|rdma|tcp,rdma|None
-                    - ...
+            will get executed with force option.
+
         Returns:
```
aujjwal-redhat, line 84: NIT: returns 0 :)

aujjwal-redhat (approved):
LGTM

common/ops/gluster_ops/volume_ops.py
srijan-sivakumar: actually it should have been ret...

nishith-vihar (approved):
LGTM

Issue #214, opened 2021-04-22
Script for installing RPMs to be added in the tools


aujjwal-redhat:
The script for upgrading, installing, degrading to a version is to be added to the tools.

aujjwal-redhat:
I have some working code. Will push that.

aujjwal-redhat:
We will be moving this to a separate repo and continuing the development over that.

PR #225, opened 2021-04-27
Added the script to install uninstall rpms

aujjwal-redhat:
Added the script to install and uninstall rpms.
Need to take care of the dependencies for 9 and 10

Fixes: #214


aujjwal-redhat:
I have added the RPMs also if somebody wants to test and modify the code as per needed.

srijan-sivakumar:
I should not push rpms to a github repo. We should just have a script for creation of the rpms.

srijan-sivakumar:
> I have added the RPMs also if somebody wants to test and modify the code as per needed.

Why would they need the rpms to modify the code ?

aujjwal-redhat:
> > I have added the RPMs also if somebody wants to test and modify the code as per needed.
> 
> Why would they need the rpms to modify the code ?

For testing I have added the rpms. If it is fine for everyone to compile the rpms and then test it then no problem. I will remove the rpms.

aujjwal-redhat:
> I should not push rpms to a github repo. We should just have a script for creation of the rpms.

Cool.

tools/install_uninstall_release.py
```diff
@@ -0,0 +1,205 @@
...
+        """
+        This method shows the menu
+        and gets the choice as input from the user
+        """
+        cls.choice = {}
+
+        print(f"Select the OS to install its rpms or press q to cancel.")
+        os_types = ["Fedora-32", "Fedora-33", "Centos-8"]
+        operations = ['Install', "Uninstall", "Upgrade/Degrade"]
+        release_types = ['release-10', 'release-9']
```
nik-redhat, line 31: We don't have a release-10 yet, it should rather be release-8

aujjwal-redhat: I have compiled rpms for 10 as well. If it is not ok I will remove the option 10 and add release 8.

nik-redhat: And, how do you have compiled RPMs for release-10, when it's yet to be developed and 4 months(minimum) are left for release?

aujjwal-redhat: I went through the compile rpms doc and used the createrpms script in the glusterfs repo for the devel branch. :grimacing:

aujjwal-redhat: It is still devel so yeah it should not be right to call it release 10.

nik-redhat: And that's totally wrong. You can't just compile the RPMs from the devel branch and term it as the RPMs for the next major release, it doesn't have most of the planned updates yet.
Moreover, just a note users generally don't work with the code on the devel branch, and instead, they use an officially released-supported version.

> It is still devel so yeah it should not be right to call it release 10.

Yes.

aujjwal-redhat: Ack

srijan-sivakumar:
@aujjwal-redhat let's create a separate repository for this initiative. I see that this can be used for the upgrade testing automation scenario which we are looking to come up with.

aujjwal-redhat:
> @aujjwal-redhat let's create a separate repository for this initiative. I see that this can be used for the upgrade testing automation scenario which we are looking to come up with.

Yeah, that would be better. :)

Issue #208, opened 2021-04-20
Signal handling to exit the framework.


srijan-sivakumar:
The framework run cannot be stopped in a clean manner right now when it is executing. This needs to be handled in a proper manner with the help of signal handlers.

PR #227, opened 2021-04-27
Adding signal handler module

nishith-vihar:
Signal handler function is added for graceful exiting of
the test framework.

Fixes: #208


srijan-sivakumar:
@nishith-vihar why are we defining signal handlers in every module ?

Redant runs as a single process and signal handlers are to be defined on the process level.

nishith-vihar:
@srijan-sivakumar Ack. I think you are right . I checked it out again. I believe the signal handling has to be done in the main function of the redant alone. I will make the required changes.

core/redant_main.py
```diff
@@ -14,6 +14,8 @@
 from test_runner import TestRunner
 from result_handler import ResultHandler
 from environ import environ
+from signal_handler import signal_handler
+import signal
```
srijan-sivakumar, line 18: Signal being a standard module should come at the very beginning before the third party modules.

core/redant_main.py
```diff
@@ -58,7 +60,7 @@ def main():
 
     start = time.time()
     args = pars_args()
-
+
```
srijan-sivakumar, line 63: remove this space in the blank line.

core/signal_handler.py
```diff
@@ -0,0 +1,20 @@
+"""
+This file consists of functions for handling 
+signals. Signal handling is required for the
+graceful exit of the test framework
+"""
+
+import signal
```
srijan-sivakumar, line 7: Does this module require `signal` as import ? I don't see any function of signal being used here. Or maybe I'm wrong. Correct me if I'm wrong here.

srijan-sivakumar (changes requested):

core/redant_main.py
nishith-vihar: Ack

core/redant_main.py
nishith-vihar: Ack

core/signal_handler.py
nishith-vihar: Yes. I think you are right. It does not require signal module here

srijan-sivakumar (approved):
A crude way. But a start nonetheless. The next stage is to add cleanups in individual test cases whenever they see this exception being raised. @nishith-vihar can you verify if this exception being raised is seen by the test runs too ? If it is the case, then we can do a cleanup directly in the except block in the parent_tests, else we have to come with a signalling mechanism within the main and the child processes ( which are running the TC ).

srijan-sivakumar:
@aujjwal-redhat please review the code, I've removed the merge conflicts.

aujjwal-redhat (approved):
LGTM

srijan-sivakumar:
@nishith-vihar ...lint check..

Issue #229, opened 2021-04-28
Converting threading to multiprocessing


srijan-sivakumar:
The current implementation has a bottleneck at GIL. It is better to bypass that issue of using threads by using multiprocessing and having multiple processes. Even in that, we can have a long running processes who wait till the job queue ends rather than re-creating the thread again and again.

PR #230, opened 2021-04-28
Converting threading to multiprocessing

srijan-sivakumar:
The current implementation will have a bottleneck because of
the nature of how threads are executed in python, wherein threads
and the main process share the GIL even though the system might
boast of multiple processor cores.

To work around this bottleneck in python, multiprocessing module
can be used. Now one can argue that multiprocess will create a
separate process in the background and that this will be heavier than
a thread and also would consume more time than spinning up a new thread.

To solve this, we use the idea of worker processes taking up tasks from
a queue, hence processes are spun up once and are killed only when there
is no job in the queue.

One more point to note is that the multiprocessing requires some mechanism of
locking to prevent the main process form going ahead with its flow ( which in
our case would be the execution of the disruptive test cases ). Now for the time
being we have an incremental backoff system with sleep till all jobs are dead.

This will have to be replaced with a sophesticated locking and signalling mechanism.

Fixes: #229

### DIFF
--- a/core/test_runner.py
+++ b/core/test_runner.py
@@ -59,17 +59,14 @@ class TestRunner:
                 proc.join()
 
         for test in cls.non_concur_test:
-            # cls.test_results[test['moduleName'][:-3]] = []
             cls._run_test(test)
 
         """
-        Because of the infinitesimal delay in the Queue
-        it was found that sometimes the
-        Queue was empty while the other times.
-        Although this is not the correct way
-        to handle it but can be considered for now.
+        Because of the infinitesimal delay in value being reflected in Queue
+        it was found that sometimes the Queue which was empty had been given
+        some value, it still showed itself as empty.
+        TODO: Handle it without sleep.
         """
-        # if not bool(cls.concur_count):
         while cls.job_result_queue.empty():
             time.sleep(1)
 
@@ -81,7 +78,6 @@ class TestRunner:
         This method creates the threadlist for non disruptive tests
         """
         for test in cls.concur_test:
-            # cls.test_results[test['moduleName'][:-3]] = []
             cls.nd_job_queue.put(test)
         
     @classmethod
@@ -115,7 +111,6 @@ class TestRunner:
             print(Fore.RED + result_text)
             print(Style.RESET_ALL)
 
-        # cls.test_results[test_dict["moduleName"][:-3]].append(test_stats)
         result_value = { test_dict["moduleName"][:-3] : test_stats }
         cls.job_result_queue.put(result_value)
 
--- a/core/test_runner.py
+++ b/core/test_runner.py
@@ -55,19 +55,14 @@ class TestRunner:
                 else:
                    backoff_time += 1
 
-        while not cls.job_result_queue.empty():
-            curr_item = cls.job_result_queue.get()
-            key = list(curr_item.keys())[0]
-            value = curr_item[key]
-            cls.test_results[key].append(value)
+            for iter in range(cls.concur_count):
+                proc.join()
 
         for test in cls.non_concur_test:
             cls.test_results[test['moduleName'][:-3]] = []
             cls._run_test(test)
 
-
-
-        return cls.test_results
+        return cls.job_result_queue
 
     @classmethod
     def _prepare_thread_tests(cls):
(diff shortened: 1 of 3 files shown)


aujjwal-redhat (approved):
LGTM

srijan-sivakumar:
@aujjwal-redhat can you take this up and update how the result handler is taking in the data. I've modified the resultant being returned by the `test_runner`. For now, to prevent errors I've commented it out but you can clone this branch and add a new patch upon it to resolve this issue.

aujjwal-redhat:
> @aujjwal-redhat can you take this up and update how the result handler is taking in the data. I've modified the resultant being returned by the `test_runner`. For now, to prevent errors I've commented it out but you can clone this branch and add a new patch upon it to resolve this issue.

Yeah will do that.

aujjwal-redhat:
```bash
(redantenv) [root@dhcp43-74 redant]# python3 core/redant_main.py -c ../Utilities/conf.yaml -t tests/functional/glusterd/
Hello World
test_peer_probe_detach-Generic PASS

{'test_peer_probe_detach': {'timeTaken': 0.2863271236419678, 'volType': 'Generic', 'testResult': 'PASS'}}
Hello
test_glusterd_start_stop-Generic PASS


Total time taken by the framework: 41.61782693862915 sec
Table:
test_peer_probe_detach

|   Generic   |     PASS    | 0.2863271236419678 |
test_glusterd_start_stop

|   Generic   |     PASS    | 0.8022167682647705 |

(redantenv) [root@dhcp43-74 redant]# python3 core/redant_main.py -c ../Utilities/conf.yaml -t tests/functional/glusterd/test_glusterd_start_stop.py -sp
Hello
test_glusterd_start_stop-Generic PASS


Total time taken by the framework: 41.439769983291626 sec
Table:
test_glusterd_start_stop

|   Generic   |     PASS    | 0.7627096176147461 |
```

core/test_runner.py
```diff
@@ -36,19 +38,34 @@ def run_tests(cls):
...
+            # TODO replace incremental backup with a signalling and lock.
+            backoff_time = 0
+            while len(jobs) > 0:
+                jobs = [job for job in jobs if job.is_alive()]
+                if backoff_time == 20:
+                    time.sleep(backoff_time)
+                else:
+                   backoff_time += 1
+
+        while not cls.job_result_queue.empty():
```
srijan-sivakumar, line 58: Why do we need to do the modifications here ? The `_tests_run` populates the results into the queue and I don't see any repetition. I'd recommend using the queue in here and then doing the operations on result only in result handler.

srijan-sivakumar (commented):
Requesting this change.

core/test_runner.py
aujjwal-redhat: Ack. Made the required changes.

core/test_runner.py
```diff
@@ -36,44 +38,47 @@ def run_tests(cls):
...
+                if backoff_time == 20:
+                    time.sleep(backoff_time)
+                else:
+                   backoff_time += 1
+
+            for iter in range(cls.concur_count):
+                proc.join()
 
         for test in cls.non_concur_test:
             cls.test_results[test['moduleName'][:-3]] = []
```
srijan-sivakumar, line 62: Why are we still using test_results..can't we use the queue here ?

aujjwal-redhat: Ack

aujjwal-redhat:
@srijan-sivakumar have made the changes.

aujjwal-redhat:
Output:
```bash
Table:
test_vol_creation

|     rep     |     PASS    | 44.017176389694214 |
|     dist    |     PASS    | 50.94098424911499  |
|   dist-rep  |     PASS    | 50.95420694351196  |
test_file_creation

|     rep     |     PASS    | 43.994415283203125 |
|     dist    |     PASS    | 34.19332408905029  |
|   dist-rep  |     FAIL    | 38.58198857307434  |
test_sample

|     dist    |     FAIL    | 16.593501091003418 |
test_glusterd_start_stop

|   Generic   |     PASS    | 39.21120047569275 |
test_peer_probe_detach

|   Generic   |     FAIL    | 3.6294660568237305 |

```

srijan-sivakumar:
LGTM..taking this feature in..

Issue #231, opened 2021-04-29
Config file validation.


srijan-sivakumar:
The test run throws an exception:

```
Error: can't find config file or read data.
Traceback (most recent call last):
  File "core/redant_main.py", line 99, in <module>
    main()
  File "core/redant_main.py", line 62, in main
    param_obj = ParamsHandler(args.config_file)
  File "/root/redant/core/parsing/params_handler.py", line 20, in __init__
    self.server_config = self.config_hashmap['servers_info']
TypeError: 'NoneType' object is not subscriptable
```

On looking into the command I used to run the test run, I found that I had given a wrong path for the config file. So, it seems like there is no validation done over the presence of the config file.

PR #232, opened 2021-04-29
Check is added for existence and accessibility of config_file

nishith-vihar:
Check is added for existence and accessibility of config_file
Fixes: #231

### DIFF
--- a/core/parsing/test_parser.py
+++ b/core/parsing/test_parser.py
@@ -22,11 +22,7 @@ class Parser():
             dict: Hashmap for config file as a dictionary.
             None: None on failure.
         """
-        try:
-            configfd = open(filepath, 'r')
-            config_hashmap = yaml.load(configfd, Loader=yaml.FullLoader)
-            configfd.close()
-            return config_hashmap
-        except IOError:
-            print("Error: can\'t find config file or read data.")
-            return None
+        configfd = open(filepath, 'r')
+        config_hashmap = yaml.load(configfd, Loader=yaml.FullLoader)
+        configfd.close()
+        return config_hashmap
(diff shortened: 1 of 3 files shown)


core/redant_main.py
```diff
@@ -59,7 +60,11 @@ def main():
     start = time.time()
     args = pars_args()
 
-    param_obj = ParamsHandler(args.config_file)
+    if os.access(args.config_file, os.R_OK):
```
srijan-sivakumar, line 63: The check should be inside the paramhandler and not redant_main.

srijan-sivakumar (changes requested):

core/parsing/test_parser.py
```diff
@@ -28,5 +28,4 @@ def generate_config_hashmap(filepath: str) -> dict:
             configfd.close()
             return config_hashmap
         except IOError:
-            print("Error: can\'t find config file or read data.")
-            return None
+            raise IOError
```
srijan-sivakumar, line 31: Rather than adding a try catch here, let's just add a function which check the file exists or not. For example let's call the function 'file_accessible'. hence the line 25-31 can be converted into..

```
if not file_accessible(filepath, <permissions>):
    raise IOError
configfd = open(filepath, 'r')
config_hashmap = yaml.load(configfd, Loader=yaml.FullLoader)
configfd.close()
```

Now this IOError can be caught at the main in a try catch. So we just need to declutter a little by replacing the try-except inside params with just a `raise IOError`.

srijan-sivakumar (changes requested):

core/parsing/test_parser.py
nishith-vihar: Ack

srijan-sivakumar (approved):
LGTM

Issue #154, opened 2021-04-01
Rexe: Changing over to passwordless ssh


srijan-sivakumar:
The current flow expects password to be provided but it'd be better if passwordless ssh is made norm in the connection from the host to other servers and clients.

srijan-sivakumar:
The current decision is to take up for fedora 32 specific machines as base and to see why ssh is failing in fedora 33. @aujjwal-redhat , please take this to completion.

aujjwal-redhat:
After a certain level of experimentation, it was founded that Fedora 33 is not able to provide a green flag for public-key authentication in Paramiko. Tested the same code on Fedora 32 gave a success message.

PR #233, opened 2021-04-29
Added the passwordless ssh code in rexe

aujjwal-redhat:
Added the passwordless ssh code in the remote executioner. Now no user
name or password is required. To test the same you need to setup the
passwordless ssh on your servers.

Note: Setup passwordless ssh from one server to another server and from
one server to the same server as well.

Fixes: #154

### DIFF
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -1,10 +1,10 @@
+import os
 import random
 import concurrent.futures
 import paramiko
 import xmltodict
 import json
 from multipledispatch import dispatch
-import os
 
 class Rexe:
     def __init__(self, host_dict):
@@ -28,13 +28,9 @@ class Rexe:
         self.connect_flag = True
 
         for node in self.host_dict:
-            # to debug PARAMIKO in case of a failure uncomment the following line
-            # paramiko.common.logging.basicConfig(level=paramiko.common.DEBUG)
 
             node_ssh_client = paramiko.SSHClient()
             node_ssh_client.load_host_keys(os.path.expanduser('/root/.ssh/known_hosts'))
-            node_ssh_client.set_missing_host_key_policy(
-                paramiko.AutoAddPolicy())
             mykey = paramiko.RSAKey.from_private_key_file('/root/.ssh/id_rsa')
             try:
                 node_ssh_client.connect(
@@ -101,8 +97,6 @@ class Rexe:
             # Reconnection to be done.
             node_ssh_client = paramiko.SSHClient()
             node_ssh_client.load_host_keys(os.path.expanduser('/root/.ssh/known_hosts'))
-            node_ssh_client.set_missing_host_key_policy(
-                paramiko.AutoAddPolicy())
             mykey = paramiko.RSAKey.from_private_key_file('/root/.ssh/id_rsa')
             try:
                 node_ssh_client.connect(
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -4,7 +4,7 @@ import paramiko
 import xmltodict
 import json
 from multipledispatch import dispatch
-
+import os
 
 class Rexe:
     def __init__(self, host_dict):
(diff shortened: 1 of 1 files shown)


srijan-sivakumar:
> Added the passwordless ssh code in the remote executioner. Now no user
> name or password is required. To test the same you need to setup the
> passwordless ssh on your servers.
> 
> Note: Setup passwordless ssh from one server to another server and from
> one server to the same server as well.
Do you mean to say that the keys should be shared between all servers ? Like for instance we have servera, serverb and serverc along with a control_machine where we run the test framework, so do we need to setup passwordless ssh between all four machines ? If yes then why do the servers require to access the control_machine ? I can understand if we need to setup passwordless ssh between server{a, b, c} as it'd also help for geo-replication, but not the scenario where anyone from server{a, b, c} can access control_machine.

> Fixes: #154
> 
> Signed-off-by: Ayush Ujjwal [aujjwal@redhat.com](mailto:aujjwal@redhat.com)

common/rexe.py
```diff
@@ -4,7 +4,7 @@
 import xmltodict
 import json
 from multipledispatch import dispatch
-
+import os
```
srijan-sivakumar, line 7: The standards dictate that we put imports of standard modules at the very top. Hence os will come before xmltodict which is a third party module.

common/rexe.py
```diff
@@ -28,18 +28,25 @@ def establish_connection(self):
         self.connect_flag = True
 
         for node in self.host_dict:
+            # to debug PARAMIKO in case of a failure uncomment the following line
+            # paramiko.common.logging.basicConfig(level=paramiko.common.DEBUG)
```
srijan-sivakumar, line 32: Again. Not a good practice. We should have this setting in the config or in command line flag. Not as a comment and tell user to comment/uncomment it.

common/rexe.py
```diff
@@ -28,18 +28,25 @@ def establish_connection(self):
...
+
             node_ssh_client = paramiko.SSHClient()
+            node_ssh_client.load_host_keys(os.path.expanduser('/root/.ssh/known_hosts'))
             node_ssh_client.set_missing_host_key_policy(
                 paramiko.AutoAddPolicy())
```
srijan-sivakumar, line 37: Do we still need to keep AutoAddPolicy ? Even after setting the keys ? This is actually a security threat. See if this can be changed.

common/rexe.py
```diff
@@ -93,13 +100,14 @@ def execute_command(self, cmd, node):
         except Exception as e:
             # Reconnection to be done.
             node_ssh_client = paramiko.SSHClient()
+            node_ssh_client.load_host_keys(os.path.expanduser('/root/.ssh/known_hosts'))
             node_ssh_client.set_missing_host_key_policy(
                 paramiko.AutoAddPolicy())
+            mykey = paramiko.RSAKey.from_private_key_file('/root/.ssh/id_rsa')
             try:
                 node_ssh_client.connect(
                     hostname=node,
```
srijan-sivakumar, line 109: is it necessary to give `hostname=node` and `pkey=mkey` ? Just curious.

srijan-sivakumar (changes requested):

aujjwal-redhat:
> > Added the passwordless ssh code in the remote executioner. Now no user
> > name or password is required. To test the same you need to setup the
> > passwordless ssh on your servers.
> > Note: Setup passwordless ssh from one server to another server and from
> > one server to the same server as well.
> 
> Do you mean to say that the keys should be shared between all servers ? Like for instance we have servera, serverb and serverc along with a control_machine where we run the test framework, so do we need to setup passwordless ssh between all four machines ? If yes then why do the servers require to access the control_machine ? I can understand if we need to setup passwordless ssh between server{a, b, c} as it'd also help for geo-replication, but not the scenario where anyone from server{a, b, c} can access control_machine.
> 
> > Fixes: #154
> > Signed-off-by: Ayush Ujjwal [aujjwal@redhat.com](mailto:aujjwal@redhat.com)

You are right @srijan-sivakumar but I meant only the servers not the servers to the control machine. In that case, obviously, the servers can get access to the same and can cause damage to the framework as well.

common/rexe.py
aujjwal-redhat: Oh ok. Will make the change.

common/rexe.py
aujjwal-redhat: Yeah true.

common/rexe.py
aujjwal-redhat: Yeah actually I was going to ask the same. Will remove that as it is not secure.

common/rexe.py
aujjwal-redhat: Yeah. Paramiko needs the node IP address in hostname and pkey for privatekey from RSA file. If we want to pass the address directly we would need key_filename.

common/rexe.py
srijan-sivakumar: @aujjwal-redhat does it work without the AutoAddPolicy() ? has this been tested after removing host keys and retrying passwordless ssh ?

common/rexe.py
srijan-sivakumar: @aujjwal-redhat does it work without the AutoAddPolicy() ? has this been tested after removing host keys and retrying passwordless ssh ?

aujjwal-redhat: Yeah @srijan-sivakumar , it works without AutoAddPolicy().
If we remove the `load_host_keys` it does not work. The host keys are needed to authenticate.

aujjwal-redhat: It will throw this error
```bash
paramiko.ssh_exception.SSHException: Server '10.xx.xx.74' not found in known_hosts
```

srijan-sivakumar: cool. I was actually checking about the AutoAddPolicy..anyway we'd be needing the known_hosts data for passwordless ssh to work.

srijan-sivakumar:
LGTM.

srijan-sivakumar (commented):
LGTM

Issue #237, opened 2021-05-03
Adding total time taken by the framework in the result file


aujjwal-redhat:
It was found that the result file is not storing the total time taken by the framework. That can give anyone a wrong idea about the total time if in any case, the person adds all the total time taken by each test to calculate the total time. Moreover, the total time is not even displayed on the CLI in the case of the result file option. Adding that information is crucial to prevent false assumptions and calculations.

PR #238, opened 2021-05-03
Added the total time information in the result file

aujjwal-redhat:
As mentioned in the issue, the total time was not getting stored in the
result file. Hence added the same.

Fixes: #237

### DIFF
--- a/core/result_handler.py
+++ b/core/result_handler.py
@@ -1,9 +1,9 @@
 """
-This component is dedicated to
-displaying the result of the tests
-in the form of tables for better
-understanding of the performance of
-framework
+This component handles the
+results in two ways:
+
+1. display on the CLI
+2. store in a file
 """
 from prettytable import PrettyTable
 from colorama import Fore, Style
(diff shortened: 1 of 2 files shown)


core/result_handler.py
```diff
@@ -43,36 +45,42 @@ def _get_output(cls, test_results: dict, colorify: bool):
 
             cls.result += (str(table)+"\n")
 
+        cls.result += (f"\nTotal time taken by the framework is {total_time}\n")
```
srijan-sivakumar, line 48: Instead of using the string2_var += string1_var, use the formatted string for concatenation, For example,
string2_var = f"{string2_var}{string1_var}". The reason being in python every string is an object which is immutable and everytime we do an operation on it or say modify it, we'd be creating a new object which points to a new memory location to store the value. Using formatted string will prevent the intermediary memory creation which might happen multiple times if we are concatenating multiple string values.

srijan-sivakumar: Probably even in case of (str(table), see if we can add it inside the formatting string rather than doing a addition to an existing string.

srijan-sivakumar (changes requested):

core/result_handler.py
aujjwal-redhat: Ack

srijan-sivakumar (approved):
LGTM

Issue #234, opened 2021-05-01
Result handler flag with -xls flag to give output as an excel/spread sheet


srijan-sivakumar:
This is specifically important for sending out emails as it'd be easy to filter and see the output when it is an spreadsheet instead of being a simple log file.

PR #241, opened 2021-05-03
Added option for storing results in excel sheet

aujjwal-redhat:
Added an option to store the results in excel sheet. Right now, it
displays the basic test stats but later we can also add memory and other
detailed stats as well.

Fixes: #234


aujjwal-redhat:
Screenshot from 2021-05-03 13-25-17
This is the excel sheet result now.

srijan-sivakumar (commented):
LGTM :p...Should've added it before.

Issue #240, opened 2021-05-03
Removing node credentials from config file


nishith-vihar:
As the passwordless ssh is incorporated in the framework the node credentials in the config file becomes redundant and has to be removed accordingly.

PR #242, opened 2021-05-03
User credentials are removed from config file

nishith-vihar:
username and password for the nodes are removed from the config file
as they are redundant after incorporating passwordless ssh into the
test framework.

Fixes: #240

### DIFF
--- a/config/CONFIG_README.md
+++ b/config/CONFIG_README.md
@@ -6,39 +6,28 @@ The components of the config.yml file are as follows:
 
 <h3>1. servers_info</h3>
 'servers_info' is info about each server in the cluster.<br>
-Each server is defined by its ip address.
-Each server should contain 4 attributes:<br>
-1) ip(key representing server): ip address of the server.<br>
-2) brick_root: the list of directories where bricks have to be created.<br>
-3) user: the username of the server for ssh connection.<br>
-4) passwd: the password of the server for ssh connection.<br>
-All the above attributes have to defined by the user.<br>
+Each server is defined by its ip address which acts as a key representing the server.<br>
+Each server should contain 1 attributes:<br>
+1) brick_root: the list of directories where bricks have to be created.<br>
+The above attribute has to be defined by the user.<br>
 If a new server has to added, then it has to follow the convention of the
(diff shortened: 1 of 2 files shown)


config/CONFIG_README.md
```diff
@@ -6,39 +6,28 @@ The components of the config.yml file are as follows:
...
 'servers_info' is info about each server in the cluster.<br>
-Each server is defined by its ip address.
-Each server should contain 4 attributes:<br>
-1) ip(key representing server): ip address of the server.<br>
-2) brick_root: the list of directories where bricks have to be created.<br>
-3) user: the username of the server for ssh connection.<br>
-4) passwd: the password of the server for ssh connection.<br>
-All the above attributes have to defined by the user.<br>
+Each server is defined by its ip address which acts as a key representing the server.<br>
+Each server should contain 1 attributes:<br>
```
aujjwal-redhat, line 10: NIT: attribute :)

aujjwal-redhat (approved):
Looks good.

config/CONFIG_README.md
nishith-vihar: Ack

config/config.yml
```diff
@@ -6,29 +6,17 @@
...
-        user: "root"
-        passwd: "redhat"
     "4.4.4.4":
         brick_root: ["/bricks"]
-        user: "root"
-        passwd: "redhat"
         
 #clients_info - All the relevant information about the clients
 clients_info:
     "5.5.5.5":
```
srijan-sivakumar, line 18: Can we keep the client IPs as values of list ?

For example, clients_info: ["5.5.5.5", "6.6.6.6"]

srijan-sivakumar (changes requested):

config/config.yml
nishith-vihar: It would be helpful if we need to add any attribute later. If we make it into a list and need to add at least one attribute we would need to change some parts of rexe or any other code. That would be a hassle

srijan-sivakumar (approved):
LGTM

Issue #236, opened 2021-05-02
Framework failure on server shutdown in beginning.


srijan-sivakumar:
If any of the said servers or clients mentioned in the config file is shutdown or connection fails  then it'd lead to failure in the execution. Hence add a validation at the very beginning to see if there's a connection failure.

PR #243, opened 2021-05-03
Added message to handle connection failure

aujjwal-redhat:
Previously, on shutdown or connection failure a stact traceback was
displayed. Instead, a simple message is being displayed in its place
now.

Fixes: #236

### DIFF
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -36,15 +36,14 @@ class Rexe:
                 node_ssh_client.connect(
                     hostname=node,
                     pkey=mykey,
+                    timeout=15,
                     )
 
                 self.logger.debug(f"SSH connection to {node} is successful.")
             except Exception as e:
                 self.logger.error(f"Connection failure. Exception : {e}")
                 self.connect_flag = False
-                print(f"Connection failed on {node}.\nTake a look in the main.log file for more details")
-                exit(0)
-                # raise e
+                raise e
             self.node_dict[node] = node_ssh_client
 
     def deconstruct_connection(self):
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -42,7 +42,9 @@ class Rexe:
             except Exception as e:
                 self.logger.error(f"Connection failure. Exception : {e}")
                 self.connect_flag = False
-                raise e
+                print(f"Connection failed on {node}.\nTake a look in the main.log file for more details")
+                exit(0)
+                # raise e
             self.node_dict[node] = node_ssh_client
 
     def deconstruct_connection(self):
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -18,7 +18,7 @@ class Rexe:
         """
         return random.choice(list(self.node_dict.keys()))
 
-    def establish_connection(self):
+    def establish_connection(self, timeout=15):
         """
         Function to establish connection with the given
         set of hosts.
@@ -36,7 +36,7 @@ class Rexe:
                 node_ssh_client.connect(
                     hostname=node,
                     pkey=mykey,
-                    timeout=15,
+                    timeout=timeout,
                     )
 
                 self.logger.debug(f"SSH connection to {node} is successful.")
--- a/core/environ.py
+++ b/core/environ.py
@@ -1,5 +1,7 @@
 import sys
 sys.path.insert(1, ".")
+import paramiko
+from socket import timeout
 from common.mixin import RedantMixin
 
 class environ:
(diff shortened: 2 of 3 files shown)


common/rexe.py
```diff
@@ -42,7 +42,9 @@ def establish_connection(self):
             except Exception as e:
                 self.logger.error(f"Connection failure. Exception : {e}")
                 self.connect_flag = False
-                raise e
+                print(f"Connection failed on {node}.\nTake a look in the main.log file for more details")
```
srijan-sivakumar, line 45: Ahh..No @aujjwal-redhat , We should actually put a connection failure check in the environ class itself. And then stop the framework run. This print statement inside the rexe won't be good. Probably we can have a check as to what type of exception we have and if it related to node being not up or not able to connect to it in the environ's modules itself. But the print here is not the best way to go ahead.

srijan-sivakumar (changes requested):

common/rexe.py
aujjwal-redhat: Yeah, that would be better.

aujjwal-redhat:
```bash
(redantenv) [root@servera redant]# python3 core/redant_main.py -c ../Utilities/conf.yaml -rf "/home/MyWork/results.txt" -t tests/functional/glusterd/

            Oops! There was a timeout connecting the servers.
            Message:timed out
            Check and run again.
            
(redantenv) [root@servera redant]# python3 core/redant_main.py -c ../Utilities/conf.yaml -rf "/home/MyWork/results.txt" -t tests/functional/glusterd/

            It seems one of the nodes is down. 
            Message: [Errno None] Unable to connect to port 22 on 10.70.43.115.
            Check and run again.
            
```

common/rexe.py
```diff
@@ -36,6 +36,7 @@ def establish_connection(self):
                 node_ssh_client.connect(
                     hostname=node,
                     pkey=mykey,
+                    timeout=15,
```
srijan-sivakumar, line 39: Good that we have finally got to this thing. This should be a configurable parameter to be handled by the user. We can add a default value to be used if the user doesn't want to provide one. Also is this timeout reflective of connection timeout or even when we execute commands ?

srijan-sivakumar (changes requested):

common/rexe.py
aujjwal-redhat: This is just for connection timeout. So if the server doesn't respond within the timeout, the connection gets terminated.

aujjwal-redhat: > Good that we have finally got to this thing. This should be a configurable parameter to be handled by the user. We can add a default value to be used if the user doesn't want to provide one. Also is this timeout reflective of connection timeout or even when we execute commands ?

Yeah we can do that. Should we provide the option in the config file or in the arguments?

aujjwal-redhat: @srijan-sivakumar I have made the timeout a parameter that we can get from the config and pass it on. The user can configure the timeout now.

srijan-sivakumar: @aujjwal-redhat , yes that would be a good idea. To get it from config and pass it on to here and other places.

aujjwal-redhat: Yeah have added the changes in the prev commit

core/environ.py
```diff
@@ -15,7 +17,35 @@ def __init__(self, param_obj, log_path : str, log_level : str):
...
-        self.redant.establish_connection()
+        try:
+            self.redant.establish_connection()
+        except paramiko.ssh_exception.NoValidConnectionsError as e:
+            print(f'''
+            It seems one of the nodes is down. 
+            Message: {e}.
+            Check and run again.
+            ''')
+            exit(0)
```
srijan-sivakumar, line 28: I might've missed this before..but we shouldn't exit inside the rexe. The exception should be handled inside environ and then we should exit there. Because if we add exit here, then TC framework run would be disrupted. The reason as to why we need these logging is to get details as to what happened. And this being the preliminary check at the very beginning of the framework's run, i.e. inside environ, we can be sure of the nature of the underlying nodes. Also, lets not print here. Just raise exception.

srijan-sivakumar (changes requested):

core/environ.py
aujjwal-redhat: This isn't inside rexe. It's inside environ. Moreover, we are already raising the exception from the `establish_connection` in rexe. I am not sure but again raising an exception here will be just overdoing exception handling coz the environ.py file is the one taking care of the environment so the exception from the very beginning must be handled there only. Just a suggestion. :)

srijan-sivakumar: Ahh..my bad then. I guess we are good to go then.

srijan-sivakumar (approved):

srijan-sivakumar:
@nishith-vihar could you review this flow ? I've resolved the merge conflicts.

Issue #239, opened 2021-05-03
Abstract Level exception not being caught at the parent test


srijan-sivakumar:
The workers are supposed to raise an exception and go ahead with their work, but this seems like a failure wherein they're stuck in a state waiting for the next data to come from the queue even when there isn't one!

```
Process Process-3:
Traceback (most recent call last):
  File "/usr/lib64/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib64/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/root/redant/core/test_runner.py", line 125, in _worker_process
    cls._run_test(job_data, True)
  File "/root/redant/core/test_runner.py", line 94, in _run_test
    runner_thread_obj = RunnerThread(tc_class, cls.param_obj,
  File "/root/redant/core/runner_thread.py", line 17, in __init__
    self.tc_obj = tc_class(mname, param_obj, volume_type,
  File "./tests/parent_test.py", line 52, in __init__
    self.redant.volume_create(self.vol_name, self.server_list[0],
  File "./common/ops/gluster_ops/volume_ops.py", line 118, in volume_create
    ret = self.execute_abstract_op_node(cmd, node)
  File "./common/ops/abstract_ops.py", line 33, in execute_abstract_op_node
    raise Exception(ret['error_msg'])
Exception: volume create: test_vol_creation-dist-rep: failed

Process Process-1:
Traceback (most recent call last):
  File "/usr/lib64/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib64/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/root/redant/core/test_runner.py", line 125, in _worker_process
    cls._run_test(job_data, True)
  File "/root/redant/core/test_runner.py", line 94, in _run_test
    runner_thread_obj = RunnerThread(tc_class, cls.param_obj,
  File "/root/redant/core/runner_thread.py", line 17, in __init__
    self.tc_obj = tc_class(mname, param_obj, volume_type,
  File "./tests/parent_test.py", line 52, in __init__
    self.redant.volume_create(self.vol_name, self.server_list[0],
  File "./common/ops/gluster_ops/volume_ops.py", line 118, in volume_create
    ret = self.execute_abstract_op_node(cmd, node)
[...]

PR #244, opened 2021-05-03
Gluster Operations shifted from parent test constructor to parent_test_run

nishith-vihar:
Gluster Operations are shifted from constructor of parent test to
parent_test_run under the try-except clause

Fixes: #239

### DIFF
--- a/tests/parent_test.py
+++ b/tests/parent_test.py
@@ -1,7 +1,7 @@
 import traceback
 import abc
-from common.mixin import RedantMixin
 from datetime import datetime
+from common.mixin import RedantMixin
 
 
 class ParentTest(metaclass=abc.ABCMeta):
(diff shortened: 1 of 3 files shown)


tests/parent_test.py
```diff
@@ -82,6 +66,21 @@ def parent_run_test(self):
         ============================================================
         ''')
         try:
+            if not thread_flag:
+                self.redant.start_glusterd()
+                self.redant.create_cluster(self.server_list)
```
srijan-sivakumar, line 71: LGTM but we also have terminate function. Wherein we are working with glusterfs specific commands which when not handled inside a try-except block will wreck havoc. Imagine `volume_create`, that'd lead to whatever operation specified inside the `terminate` to fail. Let's rethink this approach of having a teardown. If we say teardown and setup are one of the costlier operations in test case, then we might need to do away with them.

A cleanup function is required to handle the operations of disruptive test cases smoothly but we don't need to have a stub of teardown for that. 

So let's take in the operations inside the terminate, except `deconstruct_connection` inside.

srijan-sivakumar (changes requested):
Just shifting of the operations from `terminate` to `parent_run_test`. Nice that you noticed the extra param being passed without use in the calls :)

tests/parent_test.py
nishith-vihar: Ack

srijan-sivakumar (approved):

Issue #245, opened 2021-05-03
Add information about flags in the readme file


aujjwal-redhat:
There are a lot of flags in the framework and hence there needs to be a proper documentation explaining each flag.

PR #247, opened 2021-05-04
Added information about the flags

aujjwal-redhat:
Added information about the flags.
Fixes: #245 

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,18 @@
 
 Design Doc Link : [Gluster-test Design-doc](https://docs.google.com/document/d/1D8zUSmg-00ey711gsqvS6G9i_fGN2cE0EbG4u1TOsaQ/edit?usp=sharing)
 
-### Structure:
+# Contents
+* [Structure](#structure)
+* [Set up](#set-up)
+* [About](#flags)
+
+## Structure:
 
 core: contains the core redant framework which includes parsing,test_list_builder,test_runner,runner_thread and redant_main.<br>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,3 +62,15 @@ For example,
...
+
+## About
+
+### Flags
+
+* -c, --config : Stores the path of the config file(s) to read. You need to provide the path else by default it is `None`. Moreover, this is a required argument so you need to provide it for sure.
+* -t, --test-dir : The path of the test directory where test cases exist. You can also provide the path to the specific test file. But in that case remember the `-sp` flag :upside_down_face:. This is also a required argument so don't forget it.
+*   -l, --log-dir : It stores the path of the log directory where you want the log files to be kept. By default it stores `/tmp/redant` and it is not a required argument.
+* -ll, --log-level : The log level you want for the execution.By default the log level is `I` (INFO). There are other log levels also like `D`(DEBUG), `W`(WARN) etc.
+* -cc, --concurrency-count : It stores the number of concurrent tests run. By default it is 4.
```
srijan-sivakumar, line 74: Just curious, can we use both `rf` and `xls` together :)

aujjwal-redhat: Yeah, we can. It will be actually good coz later if we want to show graphs and charts regarding the test's performance and other such stuff we can add that to excel. And the standard results will still be part of the result file.

srijan-sivakumar: cool.

Issue #248, opened 2021-05-04
Add asynchronous execution in rexe


aujjwal-redhat:
Add asynchronous execution in rexe to execute commands in the background and asynchronously. This will be used for IO commands execution. The function has to take care of the result(stdout as well as stderr) by storing it in a file for the specific process id.

srijan-sivakumar:
Even in case of the synchronous execution, we are supposed to wait till the command returns a proper status or ready to return status which is not being done!!

PR #252, opened 2021-05-05
Adding async capabilities in rexe.

srijan-sivakumar:
The current set of features don't support async behaviour
 and for the ops and tests to grow, this is a natural
extension. Async uses the same set of base libraries
as that of sync command execution.

Fixes: #248

### DIFF
--- a/common/rexe.py
+++ b/common/rexe.py
@@ -1,10 +1,12 @@
 import os
+import time
 import random
 import concurrent.futures
 import paramiko
 import xmltodict
 import json
 from multipledispatch import dispatch
+
 
 class Rexe:
     def __init__(self, host_dict):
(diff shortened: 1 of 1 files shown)


aujjwal-redhat (approved):
LGTM

common/rexe.py
```diff
@@ -129,6 +133,111 @@ def execute_command(self, cmd, node):
...
+        Returns:
+            Bool : True if the operations is completed or else False.
+        """
+        return async_obj["stdout"].channel.exit_status_ready()
+
+    def collect_async_result(self, async_obj: dict) -> dict:
+        """
+        Collect the asynch command's execution result after it ends.
+        Returns:
+            dict: Returns the resultant dictionary
```
nishith-vihar, line 202: I guess the format of the dict can be added for better understanding.

common/rexe.py
```diff
@@ -129,6 +133,111 @@ def execute_command(self, cmd, node):
...
+        A check to see if the async execution of a command which
+        was dispatched has been finished.
+        Returns:
+            Bool : True if the operations is completed or else False.
+        """
+        return async_obj["stdout"].channel.exit_status_ready()
+
+    def collect_async_result(self, async_obj: dict) -> dict:
+        """
+        Collect the asynch command's execution result after it ends.
```
nishith-vihar, line 200: nit: asynch -> async :p

nishith-vihar: The argument section can also be added with one line desc of arg.

common/rexe.py
```diff
@@ -129,6 +133,111 @@ def execute_command(self, cmd, node):
...
+        """
+        async_obj = {}
+
+        if not self.connect_flag:
+            return async_obj
+        try:
+            _, stdout, stderr = self.node_dict[node].exec_command(cmd)
+            async_obj = {"cmd": cmd, "node": node, "stdout": stdout,
+                         "stderr": stderr}
+        except Exception as e:
```
nishith-vihar, line 168: as 'e' can be ommited if not used. Correct me if its used anywhere.

nishith-vihar (approved):

Issue #260, opened 2021-05-06
Add arbiter volume option in volume ops


aujjwal-redhat:
Add arbiter option in volume ops. Volume create has replica, dist_replica

PR #261, opened 2021-05-06
Added arbiter option in volume create

aujjwal-redhat:
In volume ops, inside volume create added arbiter option to take care of
arbiter volume types.

Fixes: #260

### DIFF
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -105,11 +105,12 @@ class VolumeOps(AbstractOps):
         if "replica_count" in conf_hash:
             mul_fac = conf_hash["replica_count"]
 
+            if "arbiter_count" in conf_hash:
+                mul_fac += conf_hash["arbiter_count"]
+                
             if "dist_count" in conf_hash:
                 mul_fac *= conf_hash["dist_count"]
 
-            if "arbiter_count" in conf_hash:
-                mul_fac += conf_hash["arbiter_count"]
 
         elif "dist_count" in conf_hash:
             mul_fac = conf_hash["dist_count"]
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -108,7 +108,7 @@ class VolumeOps(AbstractOps):
             if "dist_count" in conf_hash:
                 mul_fac *= conf_hash["dist_count"]
 
-            if "arbiter_count" in conf_hash["arbiter_count"]:
+            if "arbiter_count" in conf_hash:
                 mul_fac += conf_hash["arbiter_count"]
 
         elif "dist_count" in conf_hash:
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -104,8 +104,13 @@ class VolumeOps(AbstractOps):
         cmd = ""
         if "replica_count" in conf_hash:
             mul_fac = conf_hash["replica_count"]
+
             if "dist_count" in conf_hash:
                 mul_fac *= conf_hash["dist_count"]
+
+            if "arbiter_count" in conf_hash["arbiter_count"]:
+                mul_fac += conf_hash["arbiter_count"]
+
         elif "dist_count" in conf_hash:
             mul_fac = conf_hash["dist_count"]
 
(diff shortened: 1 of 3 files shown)


common/ops/gluster_ops/volume_ops.py
```diff
@@ -104,8 +104,13 @@ def volume_create(self, volname: str, node: str, conf_hash: dict,
         cmd = ""
         if "replica_count" in conf_hash:
             mul_fac = conf_hash["replica_count"]
+
             if "dist_count" in conf_hash:
                 mul_fac *= conf_hash["dist_count"]
+
+            if "arbiter_count" in conf_hash["arbiter_count"]:
```
nik-redhat, line 111: Ig it should be only `conf_hash`?

nik-redhat (commented):
Why aren't we using the same volume ops that we have in glusto itself? They have handled all possible corner cases.

nik-redhat:
Also, try adding 1 test case related to the update being done in the ops library, this will help in testing the feature as well.

srijan-sivakumar:
> Why aren't we using the same volume ops that we have in glusto itself? They have handled all possible corner cases.

The reason being they have a hardcoded mechanism of volume type and the server and bricks to be used which we don't want to do. Coming to the logic as to how the arbiter volume is to be created, yes that can be understood and seen if it can be taken up from glusto.

common/ops/gluster_ops/volume_ops.py
srijan-sivakumar: Yes..it shouldn't be conf_hash["arbiter_count"]

srijan-sivakumar (changes requested):

srijan-sivakumar:
> Also, try adding 1 test case related to the update being done in the ops library, this will help in testing the feature as well.

@aujjwal-redhat , add arb and dist-arb in the test cases wherein we have already mentioned the volume types.

nik-redhat:
> > Why aren't we using the same volume ops that we have in glusto itself? They have handled all possible corner cases.
> 
> The reason being they have a hardcoded mechanism of volume type and the server and bricks to be used which we don't want to do. Coming to the logic as to how the arbiter volume is to be created, yes that can be understood and seen if it can be taken up from glusto.

I meant the logic only. That would involve less time and hard work and properly cover the corner cases.

common/ops/gluster_ops/volume_ops.py
aujjwal-redhat: Oh yeah. :)

aujjwal-redhat:
> > Also, try adding 1 test case related to the update being done in the ops library, this will help in testing the feature as well.
> 
> @aujjwal-redhat , add arb and dist-arb in the test cases wherein we have already mentioned the volume types.

Yeah.

aujjwal-redhat:
> Also, try adding 1 test case related to the update being done in the ops library, this will help in testing the feature as well.

Yeah true

srijan-sivakumar:
> > > Why aren't we using the same volume ops that we have in glusto itself? They have handled all possible corner cases.
> > 
> > 
> > The reason being they have a hardcoded mechanism of volume type and the server and bricks to be used which we don't want to do. Coming to the logic as to how the arbiter volume is to be created, yes that can be understood and seen if it can be taken up from glusto.
> 
> I meant the logic only. That would involve less time and hard work and properly cover the corner cases.

True, but if one were to look the volume create flow in glusto it is very straightforward as it directly takes values from config, even the brick values which actually isn't attempted here.

aujjwal-redhat:
@srijan-sivakumar lemme know if it is fine with the test case part

aujjwal-redhat:
Screenshot from 2021-05-06 20-22-46

srijan-sivakumar:
LGTM. I've tested it. One of the observations I had is that in a 2 GB RAM VM, the concurrency limit has to be set to 2 or else the VMs go for a toss. This is something we have a limitation from glusterfs side. This shows that if I'm developing TCs in one server, heavy TCs will bring down the whole system. Hence we need sanity checks for that. But even before we venture there, we need to understand that there is a limitation on how many volumes can be run in the VMs we have. Had these been bare metal servers, we can increase the concurrency limit. So, if anyone is to try the framework, DO NOT KEEP CONCURRENCY LIMIT AS DEFAULT ON THE GO> REDUCE IT!

srijan-sivakumar (approved):

aujjwal-redhat:
> LGTM. I've tested it. One of the observations I had is that in a 2 GB RAM VM, the concurrency limit has to be set to 2 or else the VMs go for a toss. This is something we have a limitation from glusterfs side. This shows that if I'm developing TCs in one server, heavy TCs will bring down the whole system. Hence we need sanity checks for that. But even before we venture there, we need to understand that there is a limitation on how many volumes can be run in the VMs we have. Had these been bare metal servers, we can increase the concurrency limit. So, if anyone is to try the framework, DO NOT KEEP CONCURRENCY LIMIT AS DEFAULT ON THE GO> REDUCE IT!

ACK

Issue #263, opened 2021-05-06
Readme, design decisions and thought process


srijan-sivakumar:
The architects of any project won't be there forever with it ( not everyone has the luxury to be a BDFL ), hence it is important to have the thought process documented so that one doesn't need to go through the code. I for one believe in proper documentation. The very ideal of developers and engineers being spartans who understand logic only from code is what I feel as BS. We need to be civilized humans and make it easy for the next person coming in to just glance at what it is, why it is and how it is.

PR #264, opened 2021-05-06
Readme, design decisions and thought process

srijan-sivakumar:
The architects of any project won't be there forever with it
( not everyone has the luxury to be a BDFL ), hence it is
important to have the thought process documented so that one
doesn't need to go through the code. I for one believe in proper
documentation. The very ideal of developers and engineers being
spartans who understand logic only from code is what I feel as BS.
We need to be civilized humans and make it easy for the next person
coming in to just glance at what it is, why it is and how it is.

Fixes: #263


srijan-sivakumar:
@aujjwal-redhat @nishith-vihar @nik-redhat , this is the initial level of documentation from the perspective of the framework. More to come. But I guess for now this is enough. Will keep on adding from time to time.

nik-redhat (approved):
lgtm.
On a side note, @srijan-sivakumar have you ever thought of becoming an author or publishing a book? :stuck_out_tongue_winking_eye:

Issue #268, opened 2021-05-07
Success messages not in logs


aujjwal-redhat:
```bash
10.70.43.74:/glusterfs/brick/test_add_brick-rep-0 10.70.43.115:/glusterfs/brick/test_add_brick-rep-1 10.70.43.74:/glusterfs/brick/test_add_brick-rep-2 force on 10.70.43.74
[2021-05-07 16:16:02,798] INFO  - Running gluster vol add-brick test_add_brick-rep replica 4 10.70.43.115:/glusterfs/brick/test_add_brick-rep-3 --xml force on 10.70.43.74
.
.
       ============================================================
        test_add_brick-rep-rep Finished Running: 2021-05-07 16:16:09.930813
        ============================================================
```
Success messages not added in abstract ops.

aujjwal-redhat:
Success messages are removed from the framework as they need to  only show the errors. This needs to be taken care of in gluster ops as well.

Issue #250, opened 2021-05-04
Requirement of `cmd run successfully` in all operations.


srijan-sivakumar:
This seems like an overkill. Even though the improvement received might be in small units, it is but a logical step to not log for successful run of a command. The reasoning being just stating running command xyz on node or nodes xyz without showing any errors for the same command, will imply that it succeeded, hence we can eliminate these success logs.

BarakSason:
I disagree with the statement "just stating running command xyz on node or nodes xyz without showing any errors for the same command, will imply that it succeeded" - There are numerous cases in which an error can occur without actually being propagated back (simplest example is deadlock on the server which will cause prevent the command result from even returning), in these case, without specifically logging a success/failure, it'd be impossible to determine success with 100% certainty.

PR #269, opened 2021-05-07
Added success message in abstract ops

aujjwal-redhat:
Added success message in abstract ops.
Fixes: #268


nishith-vihar:
Success messages were deliberately removed. We only log failure messages and omit success messages.
You can refer to this issue: #250

aujjwal-redhat:
Oh, ok. Then we need to remove the same from gluster_ops as we are logging there those messages as well.

Issue #235, opened 2021-05-01
Peer operation failure leads to cascading failures and eventual TC framework failure


srijan-sivakumar:
Now I can understand a failure in a test case as an indication of bug in the product, but the testing framework should handle the exception and move on to the next test case rather than getting bogged down and eventually giving up. For example,

```
Traceback (most recent call last):
  File "./core/redant_main.py", line 99, in <module>
    main()
  File "./core/redant_main.py", line 87, in main
    result_queue = TestRunner.run_tests()
  File "/root/redant-files/redant/core/test_runner.py", line 61, in run_tests
    cls._run_test(test)
  File "/root/redant-files/redant/core/test_runner.py", line 94, in _run_test
    runner_thread_obj = RunnerThread(tc_class, cls.param_obj,
  File "/root/redant-files/redant/core/runner_thread.py", line 17, in __init__
    self.tc_obj = tc_class(mname, param_obj, volume_type,
  File "./tests/parent_test.py", line 48, in __init__
    self.redant.create_cluster(self.server_list)
  File "./common/ops/gluster_ops/peer_ops.py", line 221, in create_cluster
    self.delete_cluster(temp_cluster)
  File "./common/ops/gluster_ops/peer_ops.py", line 249, in delete_cluster
    self.peer_detach(nd, node)
  File "./common/ops/gluster_ops/peer_ops.py", line 66, in peer_detach
    ret = self.execute_abstract_op_node(cmd, node)
  File "./common/ops/abstract_ops.py", line 37, in execute_abstract_op_node
    raise Exception(ret['msg']['opErrstr'])
Exception: <sample_ip> is not part of cluster
```

This inturn implies that the peer ops is not handling all the possible return values after a peer probe.

srijan-sivakumar:
Resurfaced...

```
[...]

PR #271, opened 2021-05-07
PeerOps Modified

nishith-vihar:
when peer probed function is added to wait until peer connects
and function is added to check if peer is connected and few
other functions are modified.

Fixes: #235

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -23,8 +23,8 @@ jobs:
 
       - name: Lint with flake8
         run: |
-          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402 --count --select=E9,F63,F7,F82 --show-source --statistics
-          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402 --count --max-complexity=10 --max-line-length=79 --statistics
+          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402,W503 --count --select=E9,F63,F7,F82 --show-source --statistics
+          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402,W503 --count --max-complexity=10 --max-line-length=79 --statistics
       - name: Lint with Pylint
         run: |
           pylint --rcfile=$(pwd)/.pylintrc $(pwd)/core/
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -258,8 +258,8 @@ class PeerOps(AbstractOps):
                     self.logger.error(
                         f"Invalid UUID for the node '{peer_stat['hostname']}'")
                     is_connected = False
-                if (peer_stat['stateStr'] != "Peer in Cluster" or
-                        peer_stat['connected'] != '1'):
+                if (peer_stat['stateStr'] != "Peer in Cluster"
+                        or peer_stat['connected'] != '1'):
                     self.logger.error(f"Peer '{peer_stat['hostname']}' "
                                       f"not in connected state")
                     is_connected = False
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -165,6 +165,8 @@ class PeerOps(AbstractOps):
                               be represented by ip or hostname.
             node (str): The node which is represented by localhost. Has to
                         be replaced by corresponding IP.
+        Returns:
+            list : list of converted IPs
         """
         if 'localhost' in node_list:
             node_list.remove('localhost')
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -8,7 +8,6 @@ import random
 from time import sleep
 import socket
 import re
-from time import sleep
 from common.ops.abstract_ops import AbstractOps
 
 
@@ -85,7 +84,8 @@ class PeerOps(AbstractOps):
 
         Returns:
             'peer'(list|dict): If single peer is present then dict is returned
-                               If multiple peers are present then list is returned
+                               If multiple peers are present then list is
+                               returned
         """
 
         cmd = 'gluster --xml peer status'
@@ -260,8 +260,8 @@ class PeerOps(AbstractOps):
                     is_connected = False
                 if (peer_stat['stateStr'] != "Peer in Cluster" or
                         peer_stat['connected'] != '1'):
-                    self.logger.error(
-                        f"Peer '{peer_stat['hostname']}' not in connected state")
+                    self.logger.error(f"Peer '{peer_stat['hostname']}' "
+                                      f"not in connected state")
                     is_connected = False
 
         if not is_connected:
@@ -269,7 +269,8 @@ class PeerOps(AbstractOps):
 
         return True
 
-    def wait_for_peers_to_connect(self, node: str, servers: list, wait_timeout: int = 10) -> bool:
+    def wait_for_peers_to_connect(self, node: str, servers: list,
+                                  wait_timeout: int = 10) -> bool:
         """
         Checks nodes are peer connected with timeout.
         Args:
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -19,7 +19,7 @@ class PeerOps(AbstractOps):
     and list of peers in the pool.
     """
 
-    def peer_probe(self, server: str, node: str)->dict:
+    def peer_probe(self, server: str, node: str) -> dict:
         """
         Adds a new peer to the cluster
         Args:
@@ -40,14 +40,14 @@ class PeerOps(AbstractOps):
 
         ret = self.execute_abstract_op_node(cmd, node)
 
-        ret_probe = self.wait_for_peers_to_connect(node,server)
-        
+        ret_probe = self.wait_for_peers_to_connect(node, server)
+
         if not ret_probe:
             raise Exception(f"Peer {node} could not be connected.")
 
         return ret
 
-    def peer_detach(self, server: str, node: str, force: bool = False)->dict:
+    def peer_detach(self, server: str, node: str, force: bool = False) -> dict:
         """
         Detach the specified server.
         Args:
@@ -77,7 +77,7 @@ class PeerOps(AbstractOps):
 
         return ret
 
-    def get_peer_status(self, node: str)->list:
+    def get_peer_status(self, node: str) -> list:
         """
         Checks the status of the peers
         Args:
@@ -85,7 +85,7 @@ class PeerOps(AbstractOps):
 
         Returns:
             'peer'(list|dict): If single peer is present then dict is returned
-                                If multiple peers are present then list is returned 
+                                If multiple peers are present then list is returned
             {
                 'uuid': '504fdf87-805a-4e8f-a266-b2f45d76958a',
                 'hostname': 'dhcp42-209.lab.eng.blr.redhat.com',
@@ -167,7 +167,7 @@ class PeerOps(AbstractOps):
 
         return pool_list
 
-    def convert_hosts_to_ip(self, node_list: list, node: str)->list:
+    def convert_hosts_to_ip(self, node_list: list, node: str) -> list:
         """
         Redant framework works with IP addresses ( especially rexe )
         hence it makes sense to have a function to handle the conversion
@@ -237,7 +237,7 @@ class PeerOps(AbstractOps):
                 continue
             self.peer_detach(nd, node)
 
-    def is_peer_connected(self, node, servers)->bool:
+    def is_peer_connected(self, node: str, servers: list) -> bool:
         """
         Checks whether specified peers are in cluster and 'Connected' state.
         Args:
@@ -268,12 +268,14 @@ class PeerOps(AbstractOps):
         for peer_stat in peer_status_list:
             if socket.gethostbyname(peer_stat['hostname']) in server_ips:
                 if (re.match(r'([0-9a-f]{8})(?:-[0-9a-f]{4}){3}-[0-9a-f]{12}',
-                            peer_stat['uuid'], re.I) is None):
-                    self.logger.error(f"Invalid UUID for the node '{peer_stat['hostname']}'")
+                             peer_stat['uuid'], re.I) is None):
+                    self.logger.error(
+                        f"Invalid UUID for the node '{peer_stat['hostname']}'")
                     is_connected = False
                 if (peer_stat['stateStr'] != "Peer in Cluster" or
                         peer_stat['connected'] != '1'):
-                    self.logger.error(f"Peer '{peer_stat['hostname']}' not in connected state")
+                    self.logger.error(
+                        f"Peer '{peer_stat['hostname']}' not in connected state")
                     is_connected = False
 
         if not is_connected:
@@ -281,7 +283,7 @@ class PeerOps(AbstractOps):
 
         return True
 
-    def wait_for_peers_to_connect(self, node, servers, wait_timeout=10)->bool:
+    def wait_for_peers_to_connect(self, node: str, servers: list, wait_timeout: int = 10) -> bool:
         """
         Checks nodes are peer connected with timeout.
         Args:
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -85,21 +85,7 @@ class PeerOps(AbstractOps):
 
         Returns:
             'peer'(list|dict): If single peer is present then dict is returned
-                                If multiple peers are present then list is returned
-            {
-                'uuid': '504fdf87-805a-4e8f-a266-b2f45d76958a',
-                'hostname': 'dhcp42-209.lab.eng.blr.redhat.com',
-                'hostnames': {
-                                'hostname': [
-                                                'dhcp1.1.lab.eng.blr.redhat.com',
-                                                '1.1.1.1'
-                                            ]
-                            },
-                'connected': '1',
-                'state': '3',
-                'stateStr': 'Peer in Cluster'
-            }
-
+                               If multiple peers are present then list is returned
         """
 
         cmd = 'gluster --xml peer status'
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -7,7 +7,6 @@ from the test case.
 import random
 from time import sleep
 import socket
-import re
 from common.ops.abstract_ops import AbstractOps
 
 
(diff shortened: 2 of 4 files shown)


common/ops/gluster_ops/peer_ops.py
```diff
@@ -247,3 +238,72 @@ def delete_cluster(self, node_list: list):
             if nd == node:
                 continue
             self.peer_detach(nd, node)
+
+    def is_peer_connected(self, node, servers)->bool:
```
aujjwal-redhat, line 242: NIT:
node: str
servers : list

common/ops/gluster_ops/peer_ops.py
```diff
@@ -247,3 +238,72 @@ def delete_cluster(self, node_list: list):
...
+        return True
+
+    def wait_for_peers_to_connect(self, node, servers, wait_timeout=10)->bool:
+        """
+        Checks nodes are peer connected with timeout.
+        Args:
+            node: node on which cmd has to be executed.
+            servers (str|list): A server|List of server hosts on which peer
+                status has to be checked.
+            wait_timeout: timeout to retry connected status check in node.
```
aujjwal-redhat, line 293: Same as above.

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
```diff
@@ -67,54 +75,39 @@ def peer_detach(self, server: str, node: str, force: bool = False):
...
 
         Returns:
-            ret: A dictionary consisting
-                - Flag : Flag to check if connection failed
-                - msg : message
-                - error_msg: error message
-                - error_code: error code returned
-                - cmd : command that got executed
-                - node : node on which the command got executed
+            'peer'(list|dict): If single peer is present then dict is returned
```
srijan-sivakumar, line 85: @nishith-vihar , let's shift these examples into the readme. This would go to the doc/RP/common/ops/peer_ops.md

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
             if nd == node:
                 continue
             self.peer_detach(nd, node)
+
+    def is_peer_connected(self, node: str, servers: list) -> bool:
+        """
+        Checks whether specified peers are in cluster and 'Connected' state.
+        Args:
+            node (str): Node from which peer probe has to be executed.
```
nik-redhat, line 230: nit: peer status

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
...
+        peer_status_list = self.get_peer_status(node)
+        if peer_status_list is None:
+            return False
+        elif not isinstance(peer_status_list, list):
+            peer_status_list = [peer_status_list]
+
+        # Convert all hostnames to ip's
+        server_ips = []
+        for server in servers:
+            server_ips.append(socket.gethostbyname(server))
```
nik-redhat, line 251: Does this guarantee that the servers passed would be hostnames and not IP addresses?

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
...
+
+        is_connected = True
+        for peer_stat in peer_status_list:
+            if socket.gethostbyname(peer_stat['hostname']) in server_ips:
+                if (re.match(r'([0-9a-f]{8})(?:-[0-9a-f]{4}){3}-[0-9a-f]{12}',
+                             peer_stat['uuid'], re.I) is None):
+                    self.logger.error(
+                        f"Invalid UUID for the node '{peer_stat['hostname']}'")
+                    is_connected = False
```
nik-redhat, line 260: This case isn't possible as Gluster would handle the generation of UUID and if that's wrong it would report an error.

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
...
+                if (peer_stat['stateStr'] != "Peer in Cluster"
+                        or peer_stat['connected'] != '1'):
+                    self.logger.error(f"Peer '{peer_stat['hostname']}' "
+                                      f"not in connected state")
+                    is_connected = False
+
+        if not is_connected:
+            return False
+
+        return True
```
nik-redhat, line 270: Simply returning `is_connected` should do?

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
...
+            servers = [servers]
+
+        if node in servers:
+            servers.remove(node)
+
+        peer_status_list = self.get_peer_status(node)
+        if peer_status_list is None:
+            return False
+        elif not isinstance(peer_status_list, list):
+            peer_status_list = [peer_status_list]
```
nik-redhat, line 246: If we are ultimately converting the value to a list then why at all pass a dictionary? What benefit does that give?

common/ops/gluster_ops/peer_ops.py
nik-redhat, line 260: Also, if we do not `break` from the loop here and at line: 265, then the final result might be wrong considering the peer in this iteration is not connected but the next ones are and the end value is `True`, which is wrong.

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +222,75 @@ def delete_cluster(self, node_list: list):
...
+            servers = [servers]
+
+        count = 0
+        while count <= wait_timeout:
+            ret = self.is_peer_connected(node, servers)
+            if ret:
+                return True
+            sleep(1)
+            count += 1
+        self.logger.error(f"Peers are not in connected state: {servers}")
```
nik-redhat, line 295: This is not needed as the `is_peer_connected` function should log the correct result. But, that too needs to be modified a bit, check the above comment.

core/environ.py
```diff
@@ -53,8 +54,15 @@ def setup_env(self):
         """
         Setting up of the environment before the TC execution begins.
         """
-        self.redant.start_glusterd()
-        self.redant.create_cluster(self.server_list)
+        try:
+            self.redant.start_glusterd()
+            self.redant.create_cluster(self.server_list)
+        except Exception as error:
+            print(error)
```
nik-redhat, line 61: Was this added for testing or we actually need it?

nik-redhat (changes requested):

common/ops/gluster_ops/peer_ops.py
```diff
@@ -17,7 +18,7 @@ class PeerOps(AbstractOps):
     and list of peers in the pool.
     """
 
-    def peer_probe(self, server: str, node: str):
+    def peer_probe(self, server: str, node: str) -> dict:
```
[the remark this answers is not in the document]
nishith-vihar: All the ops are following the same format. If it is really not required then we can later raise an issue and take it up in a different patch for all the ops.

common/ops/gluster_ops/peer_ops.py
```diff
@@ -38,9 +39,14 @@ def peer_probe(self, server: str, node: str):
```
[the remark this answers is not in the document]
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
```diff
@@ -38,9 +39,14 @@ def peer_probe(self, server: str, node: str):
 
         ret = self.execute_abstract_op_node(cmd, node)
 
+        ret_probe = self.wait_for_peers_to_connect(node, server)
+
+        if not ret_probe:
+            raise Exception(f"Peer {node} could not be connected.")
+
         return ret
```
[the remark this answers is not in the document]
nishith-vihar: I think i have to change that

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

core/environ.py
nishith-vihar: Yeah. I think it was for testing. Will remove it

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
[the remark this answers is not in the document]
nik-redhat: If all the ops are using it then it's fine, though I don't think it helps, because by default python will simply ignore it and just useful for readability. If we actually want to do a check, we will have to use a separate library.

common/ops/gluster_ops/peer_ops.py
nishith-vihar: actually when the peer status command is executed, if there is only one node then 'dict' is returned when the output is converted to xml format back in the rexe itself. so that is why we convert that to list over here which is only in case of one node scenario. if there are multiple nodes the we return list from get_peer_status command

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Yeah. actually it is all IPs. conversion is not required here.

common/ops/gluster_ops/peer_ops.py
nik-redhat: So, are you saying it's controlled centrally for all such cases where we return a single value it would be a dictionary?

nishith-vihar: yes

common/ops/gluster_ops/peer_ops.py
nishith-vihar: That is not the case here. Even if the next node is connected, the variable is not made true.

common/ops/gluster_ops/peer_ops.py
nik-redhat: Ok, then we can ignore it.

common/ops/gluster_ops/peer_ops.py
nik-redhat: How?

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
nik-redhat: Ohh, yes got it. But, again what's the benefit in continuing with the loop if one of the peers is disconnected?

nishith-vihar: just so that we get to know all the servers which are not connected at once rather than connecting the first unconnected server just to realize after running it one more time that another server is not connected.

nik-redhat: Makes sense.

.github/workflows/linting.yml
```diff
@@ -23,8 +23,8 @@ jobs:
 
       - name: Lint with flake8
         run: |
-          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402 --count --select=E9,F63,F7,F82 --show-source --statistics
-          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402 --count --max-complexity=10 --max-line-length=79 --statistics
+          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402,W503 --count --select=E9,F63,F7,F82 --show-source --statistics
+          flake8 $(pwd) --ignore=C901,F811,F841,E722,E226,E402,W503 --count --max-complexity=10 --max-line-length=79 --statistics
```
aujjwal-redhat, line 27: Can you add the same to the script :) ?

aujjwal-redhat: tools/linting.sh

common/ops/gluster_ops/peer_ops.py
```diff
@@ -170,23 +149,22 @@ def get_pool_list(self, node: str) -> list:
...
         hence it makes sense to have a function to handle the conversion
         a node_list containing hostnames to ip addresses and if there's
         a localhost term, that is replaced by the node value.
         Args:
             node_list (list): List of nodes obtained wherein the node can
-            be represented by ip or hostname.
+                              be represented by ip or hostname.
             node (str): The node which is represented by localhost. Has to
-            be replaced by corresponding IP.
+                        be replaced by corresponding IP.
```
aujjwal-redhat, line 167: NIT: return missing in docstring

nik-redhat (changes requested):

common/ops/gluster_ops/peer_ops.py
```diff
@@ -36,16 +36,23 @@ def peer_probe(self, server: str, node: str):
 
         cmd = f'gluster --xml peer probe {server}'
 
-        ret = self.execute_abstract_op_node(cmd, node)
+        self.execute_abstract_op_node(cmd, node)
```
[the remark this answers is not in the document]
nishith-vihar: If the command execution fails here then exception is raised in abstract ops which is caught by parent test. So there will not be any need to check the return status.

.github/workflows/linting.yml
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
nishith-vihar: Ack

common/ops/gluster_ops/peer_ops.py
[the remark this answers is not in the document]
nik-redhat: Ack.

common/ops/gluster_ops/peer_ops.py
```diff
@@ -245,3 +223,61 @@ def delete_cluster(self, node_list: list):
...
+
+        peer_status_list = self.get_peer_status(node)
+        if peer_status_list is None:
+            return False
+        elif not isinstance(peer_status_list, list):
+            peer_status_list = [peer_status_list]
+
+        is_connected = True
+        for peer_stat in peer_status_list:
+            if (socket.gethostbyname(peer_stat['hostname']) in servers
```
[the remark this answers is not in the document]
nishith-vihar: Actually the ip address in the peer status is held by the 'hostname' attribute. so if it has a hostname then it will convert into ip or else it will stay as ip.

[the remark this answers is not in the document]
nik-redhat: Ohh, that doesn't matter. It will contain the value that you will supply while peer probing, if you peer probe using IP address then that `hostname` field would have the IP address only.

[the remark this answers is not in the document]
nishith-vihar: No. I think sometimes even if we give ip address and if it has a hostname it displays the hostname for the peer status command. for example the below is the output of my peer status:
`Hostname: dhcp42-209.lab.eng.blr.redhat.com
Uuid: 504fdf87-805a-4e8f-a266-b2f45d76958a
State: Peer in Cluster (Disconnected)`
even though i have never given 'dhcp42-209.lab.eng.blr.redhat.com' while probing.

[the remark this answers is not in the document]
nik-redhat: Actually, that depends upon the `/etc/hosts` file of the system. Suppose I change the reference to use the IP name in both the place it will work fine.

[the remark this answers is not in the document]
nishith-vihar: Got it. So now basically the change to make is from `socket.gethostbyname(peer_stat['hostname'])` to `peer_stat['hostname']` right?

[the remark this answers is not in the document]
srijan-sivakumar: Ok..just to add a point here.. we can do the changes in the /etc/hosts, but @nik-redhat , lets not leave it to the user. If we can handle it in here, it is better to do so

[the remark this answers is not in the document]
nik-redhat: Yes, for this code.
But at the VMs level, we still need to see how the hosts are configured so that we get the correct value.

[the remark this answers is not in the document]
srijan-sivakumar: I have experimented with this part and it turns into failure many a times and we don't want to hear the same complaints which people usually tag with Glusto, if we are taking the hostname, and that too only when we don't have the ip as response. Now even in this scenario, the tests can fail if the user has added hostname in his /etc/hosts and that won't be available when another node tries to find its resolution with the DNS. Now that is a configuration user will have to guard properly. So, either way, the user would have to add the hostname to the /etc/hosts, they cannot have it one way and only in one server. So, we can enforce one edict that don't add custom names to the servers in /etc/hosts then with the current code, even if we were to get hostname that will be a FQDN.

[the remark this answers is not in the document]
nik-redhat: I know, but I am not sure if this function could increase the time consumption as well. Since we are doing it for each server, and suppose in a test case we are doing a status check multiple times.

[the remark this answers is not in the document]
srijan-sivakumar: So, my proposal is an enforcement..I mean you cannot have the cake and eat it too :p, plus one server config missed or wrong config added in /etc/hosts can wreck havoc for our flow.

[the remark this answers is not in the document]
srijan-sivakumar: > > Ok..just to add a point here.. we can do the changes in the /etc/hosts, but @nik-redhat , lets not leave it to the user. If we can handle it in here, it is better to do so
> 
> I know, but I am not sure if this function could increase the time consumption as well. Since we are doing it for each server, and suppose in a test case we are doing a status check multiple times.

This would be a micro improvement. Also non-disruptive test cases would not be re-peering most of the times. The disruptive cases might do that, and even in those scenarios, I'm trying to come up with a way to handle complete re-peering from scratch. So I don't think a single call to the local DNS/switch for FQDN resolutions will cause much issue.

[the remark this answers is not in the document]
nik-redhat: Okay, then I would leave it to this :slightly_smiling_face:

nik-redhat (approved):

Issue #287, opened 2021-05-10
Lint script needs to be modified


aujjwal-redhat:
Lint script needs to be changed as per the latest commits pushed. This will help anyone to perform the lint operations easily.

PR #288, opened 2021-05-10
Lint script modified

aujjwal-redhat:
Lint script changed as per the recent commits pushed. Now, anyone can
test the lint by just running `tools/linting.sh -f -l` and both flake
and pylint will run and provide results similar to the check that take
place through linting.yaml

Fixes: #287


srijan-sivakumar (approved):
LGTM

Issue #282, opened 2021-05-10
Volds being accessed openly


srijan-sivakumar:
Currently, the volds is being accessed directly, but it'll make sense if we could define the functionalities corresponding to it, atleast a basic set of getters and setters.

PR #290, opened 2021-05-10
[Common] Guarding volds access.

srijan-sivakumar:
The volds data structure should be guarded with getters
and setters.

Fixes: #282


srijan-sivakumar:
@nishith-vihar @aujjwal-redhat if more getters and new setters are required, we can build on what we have in here.

aujjwal-redhat (commented):
NIT: Args and return statements in docstrings

common/mixin.py
```diff
@@ -15,3 +15,63 @@ class RedantMixin(GlusterOps, VolumeOps, PeerOps, IoOps, Rexe, Logger):
...
+            node (str) client node.
+        Returns:
+            list of mountpaths belonging to a node or list of
+            all mountpoints
+        """
+        if node is None:
+            mount_point_list = []
+            for (_, mnt_pts) in self.volds[volname]['mountpath'].items():
+                mount_point_list.append(mnt_pts)
+            return mount_point_list
```
nishith-vihar, line 47: Can we have mountpoints in the following format
[ {"client": ip_address, "mountpath": path}, {"client": ip_address, "mountpath": path}]
Basically a list of dictionaries.

nishith-vihar (changes requested):

common/mixin.py
srijan-sivakumar: The problem will be that is we have multiple mountpoints in a client, we cannot accommodate this format as duplication is not allowed in keys in a dictionary ( dictionary in python is but hash in the background ).

srijan-sivakumar: ok..if I understand you correctly, the suggestion is to change the format from

```
{"clientip" : ["mnt1", "mnt2"],
...
}
```

to 

```
[
{'clientip1' : ['mnt1', 'mnt2']},
...
]
```

nishith-vihar (approved):

srijan-sivakumar:
@aujjwal-redhat @nishith-vihar added the dependency now..

Issue #292, opened 2021-05-10
Test Conversions


srijan-sivakumar:
A stub issues which can be referred whenever we migrate TC from glusto-tests to Redant.

PR #293, opened 2021-05-10
[Test] Test glusterd set reset reserve limit

srijan-sivakumar:
The test case is concerned with reserve limit option in a volume.
Following set of operations are performed,
1. Set storage reserve limit to some value
2. validate it
3. Reset it
4. Validate if it is reset properly.

#Updates: #292
Signed-off-by: srijan-sivakumar ssivakum@redhat.com


nishith-vihar (approved):
LGTM

aujjwal-redhat (approved):
LGTM

Issue #303, opened 2021-05-11
Add test_default_log_level of CLI


aujjwal-redhat:
Migrate the test from glusto-tests and make modifications as per the functionalities in redant.

PR #304, opened 2021-05-11
Added test_default_log_level_of_cli

aujjwal-redhat:
Migrated the test case from glusto-test.
Made modifications as per the functionalities in redant.

Updates : #292
Fixes : #303

### DIFF
--- a/tests/functional/glusterd/test_default_log_level_of_cli.py
+++ b/tests/functional/glusterd/test_default_log_level_of_cli.py
@@ -1,0 +1,97 @@
+#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+""" Description:
+      Test to check that default log level of CLI should be INFO
+"""
+
+from glusto.core import Glusto as g
+from glustolibs.gluster.exceptions import ExecutionError
+from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
+from glustolibs.gluster.volume_ops import (volume_start, volume_status,
+                                           volume_info, volume_stop)
+
+
+@runs_on([['distributed', 'replicated', 'distributed-replicated',
+           'dispersed', 'distributed-dispersed', 'arbiter',
+           'distributed-arbiter'], ['glusterfs']])
+class TestDefaultLogLevelOfCLI(GlusterBaseClass):
+    def setUp(self):
+        # Calling GlusterBaseClass setUp
+        self.get_super_method(self, 'setUp')()
+
+        # Creating and starting the volume
+        ret = self.setup_volume()
+        if not ret:
+            raise ExecutionError("Volume creation/start operation"
+                                 " failed: %s" % self.volname)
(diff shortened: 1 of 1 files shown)


srijan-sivakumar:
@nik-redhat one of your test case is being migrated :p

aujjwal-redhat:
> @nik-redhat one of your test case is being migrated :p

Yeah :smiley:

srijan-sivakumar:
> I will add all the volumes and test this once

Just a note that, for now, in migration, lets just migrate with the same set of details..if the glusto-tests says test for dist, lets just test for dist. Because at the end of the day, we'd be comparing glusto-tests with redant after these additions and we wouldn't want to be seeing more than what we were supposed to. Maybe we can enhance the test case later...but for now, I think lets keep it the way it is..

aujjwal-redhat:
Screenshot from 2021-05-11 16-33-53
I had run the framework with DEBUG level and we can see it's showing the same in the logs
Screenshot from 2021-05-11 16-37-48 

Screenshot from 2021-05-11 16-37-15

tests/functional/glusterd/test_default_log_level_of_cli.py
```diff
@@ -0,0 +1,97 @@
...
+from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
+from glustolibs.gluster.volume_ops import (volume_start, volume_status,
+                                           volume_info, volume_stop)
+
+
+@runs_on([['distributed', 'replicated', 'distributed-replicated',
+           'dispersed', 'distributed-dispersed', 'arbiter',
+           'distributed-arbiter'], ['glusterfs']])
+class TestDefaultLogLevelOfCLI(GlusterBaseClass):
+    def setUp(self):
```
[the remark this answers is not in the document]
srijan-sivakumar: oh...ok..then no worried, I thought maybe it is just restricted to one particular volume type. Cool then..

tests/functional/glusterd/test_default_log_level_of_cli.py
```diff
@@ -0,0 +1,62 @@
...
+        cmd = "ls -tr /tmp/redant/ | tail -1"
+        ret = redant.execute_command(cmd, self.server_list[0])
+        log_file = ret['msg'][0][:-1]
+        cmd = f"cat /tmp/redant/'{log_file}'/main.log | grep -F DEBUG | wc -l"
+        ret = redant.execute_command(cmd, self.server_list[0])
+
+        if int(ret['msg'][0].split("\\")[0]) == 0:
+            redant.logger.info("The default log level is INFO")
+        else:
+            redant.logger.info("The default log level is DEBUG")
```
nik-redhat, line 62: The test should fail, if the default is not `INFO`.

aujjwal-redhat: Oh ok. Will raise Exception here then.

tests/functional/glusterd/test_default_log_level_of_cli.py
```diff
@@ -0,0 +1,59 @@
...
+        redant.volume_start(self.vol_name, self.server_list[0])
+
+        # Check the default log level of cli.log
+        cmd = 'cat /var/log/glusterfs/cli.log | grep -F "] D [" | wc -l'
+        ret = redant.execute_command(cmd, self.server_list[0])
+
+        if int(ret['msg'][0].split("\\")[0]) == 0:
+            redant.logger.info("The default log level is INFO")
+        else:
+            raise Exception("Default log level not INFO")
```
nik-redhat, line 59: What's the difference with using `redant.logger.error`? Will that not end the test with the error message?

aujjwal-redhat: And the issue will be the test will pass if we don't raise exception.

srijan-sivakumar:
@aujjwal-redhat @nishith-vihar , let's also add a link of the test case in glusto-tests repo with the commit message, so that it is easy to review and check if we are properly migrating.

tests/functional/glusterd/test_default_log_level_of_cli.py
nik-redhat: Okay cool. So, what will be shown on the CLI?

nik-redhat (approved):
Good that you started with my TC, now the journey ahead would be easier :stuck_out_tongue_winking_eye:

srijan-sivakumar (approved):
LGTM

Issue #305, opened 2021-05-11
Add test_default_max_brick_per_process


aujjwal-redhat:
Migrate the test from glusto tests.
Make modifications as per the functionalities of redant.

PR #307, opened 2021-05-11
Added test_default_max_bricks_per_process

aujjwal-redhat:
Migrated the [test from glusto-tests](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_default_max_bricks_per_process.py)
Made changes as per the functionalities in redant.

Updates : #292
Fixes : #305

### DIFF
--- a/tests/functional/glusterd/test_default_max_bricks_per_process.py
+++ b/tests/functional/glusterd/test_default_max_bricks_per_process.py
@@ -46,26 +46,15 @@ class TestDefaultMaxBricksPerProcess(AbstractTest):
         redant.reset_volume_option('all', 'all', self.server_list[0])
 
         # Fetch the max bricks per process value
-        ret = redant.get_volume_options(node=self.server_list[0])
-
-        # Comparing the values
-        second_value = ret['cluster.max-bricks-per-process']
-
-        if initial_value != second_value:
-            raise Exception("Unexpected: Max"
-                            " bricks per-process value is not equal")
+        redant.validate_volume_option('all',
+                                      {'cluster.max-bricks-per-process': initial_value},
+                                      self.server_list[0])
 
         # Enable brick-multiplex in the cluster
         redant.set_volume_options(
             'all', {'cluster.brick-multiplex': 'enable'},
             self.server_list[0])
 
-        # Fetch the max bricks per process value
-        ret = redant.get_volume_options(node=self.server_list[0])
-
-        # Comparing the values
-        third_value = ret['cluster.max-bricks-per-process']
-
-        if third_value != initial_value:
-            raise Exception("Unexpected: Max"
-                            " bricks per-process value is not equal")
+        redant.validate_volume_option('all',
+                                      {'cluster.max-bricks-per-process': initial_value},
+                                      self.server_list[0])
--- a/tests/functional/glusterd/test_default_max_bricks_per_process.py
+++ b/tests/functional/glusterd/test_default_max_bricks_per_process.py
@@ -1,0 +1,100 @@
+#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


tests/functional/glusterd/test_default_max_bricks_per_process.py
```diff
@@ -0,0 +1,71 @@
...
+        2) Fetch the max bricks per process value
+        3) Reset the volume options
+        4) Fetch the max bricks per process value
+        5) Compare the value fetched in last step with the initial value
+        6) Enable brick-multiplexing in the cluster
+        7) Fetch the max bricks per process value
+        8) Compare the value fetched in last step with the initial value
+        """
+        # Fetch the max bricks per process value
+        ret = redant.get_volume_options(node=self.server_list[0])
```
nik-redhat, line 42: At what level is this fetching the value?
Volume level or cluster level?

aujjwal-redhat: Cluster level. By default it is `all`.
```bash
    def get_volume_options(self, volname: str = 'all', option: str = 'all',
                           node: str = None) -> dict:
```

nik-redhat (commented):
Ok...so here we can incorporate the use of methods, since we are using the same function multiple times, that can be moved to a function.
Or, shall it be kept as it is since we are migrating the tests now and not updating them?
@srijan-sivakumar

srijan-sivakumar:
> Ok...so here we can incorporate the use of methods, since we are using the same function multiple times, that can be moved to a function.
> Or, shall it be kept as it is since we are migrating the tests now and not updating them?
> @srijan-sivakumar

No, we shouldn't be keeping everything as it is..Only in case of volume types for which the test case is to be run can be kept same, rest whatever optimizations or reductions we can do, we will have to do.

tests/functional/glusterd/test_default_max_bricks_per_process.py
```diff
@@ -0,0 +1,71 @@
...
+        initial_value = ret['cluster.max-bricks-per-process']
+
+        # Reset the volume options
+        redant.reset_volume_option('all', 'all', self.server_list[0])
+
+        # Fetch the max bricks per process value
+        ret = redant.get_volume_options(node=self.server_list[0])
+
+        # Comparing the values
```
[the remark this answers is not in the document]
nik-redhat: Oh, why didn't I used that while writing the test case :thinking: ...... and now  that I think of it, because of lack of documentation which the users won't have now :)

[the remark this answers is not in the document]
srijan-sivakumar: @aujjwal-redhat , after finishing off this test case, lets focus on documentation for the remaining time we have today.. we cannot let what happened here to happen again. :p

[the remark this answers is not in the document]
nik-redhat: That's fine.....my joke was intended for glusto though :p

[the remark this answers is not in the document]
aujjwal-redhat: Ack

tests/functional/glusterd/test_default_max_bricks_per_process.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Test Case:
+        1) Create a volume and start it.
+        2) Fetch the max bricks per process value
+        3) Reset the volume options
+        4) Fetch the max bricks per process value
+        5) Compare the value fetched in last step with the initial value
+        6) Enable brick-multiplexing in the cluster
+        7) Fetch the max bricks per process value
+        8) Compare the value fetched in last step with the initial value
```
nik-redhat, line 39: It can be re-written to highlight that both the fetch and comparison are done in the same function.

nik-redhat (approved):

tests/functional/glusterd/test_default_max_bricks_per_process.py
```diff
@@ -0,0 +1,65 @@
...
+  with this program; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+  Description:
+      Default max bricks per-process should be 250
+"""
+
+from tests.abstract_test import AbstractTest
+
+# disruptive;rep,dist,arb,disp,dist-rep,dist-arb,dist-disp
```
nik-redhat, line 24: I just checked this, it should be nonDisruptive right?

aujjwal-redhat: I had checked this with non-disruptive also but it was failing for random tests. I think so it's changing the volume options that might be the case

Issue #308, opened 2021-05-11
[Test] Migrating test_glusterd_quorum_command


srijan-sivakumar:
Glusto test : [test_glusterd_quorum_command](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_glusterd_quorum_command.py).

PR #312, opened 2021-05-11
[Test] test_glusterd_quorum_command migration

srijan-sivakumar:
Glusto test link : [test_glusterd_quorum_command](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_glusterd_quorum_command.py)

Fixes: #308

### DIFF
--- a/tests/functional/glusterd/test_glusterd_quorum_command.py
+++ b/tests/functional/glusterd/test_glusterd_quorum_command.py
@@ -24,6 +24,7 @@ from glustolibs.gluster.exceptions import ExecutionError
 from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
 from glustolibs.gluster.volume_ops import (
     set_volume_options,
+    volume_reset,
     get_volume_options)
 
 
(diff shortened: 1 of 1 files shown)


tests/functional/glusterd/test_glusterd_quorum_command.py
```diff
@@ -0,0 +1,66 @@
...
+    Test quorum cli commands in glusterd
+"""
+
+# disruptive;rep,dist,disp,arb,dist-arb,dist-rep
+
+from tests.abstract_test import AbstractTest
+
+
+class TestGlusterDQuorumCLICommands(AbstractTest):
+    """ Testing Quorum CLI commands in GlusterD """
```
nik-redhat, line 28: Considering we have already added a description above, this is not needed.

tests/functional/glusterd/test_glusterd_quorum_command.py
```diff
@@ -0,0 +1,66 @@
...
+                                        for_all: bool = False):
+        """
+        Method to handle the repetitive calls for setting quorum
+        options such as quorum ratio and quorum type.
+        """
+        vol = ""
+        if for_all:
+            vol = "all"
+        else:
+            vol = self.vol_name
```
nik-redhat, line 40: Instead of this....by looking at the usage I think the volume name can be passed directly as the arg.

nik-redhat (changes requested):

tests/functional/glusterd/test_glusterd_quorum_command.py
srijan-sivakumar: Ack.

tests/functional/glusterd/test_glusterd_quorum_command.py
srijan-sivakumar: Currently, I have added this as a class method here, but technically this method should exist inside the vol_ops. Coming to the part of providing the volname. Yes, now that I think we need to provide it.

aujjwal-redhat (approved):
LGTM

nik-redhat (approved):

Issue #313, opened 2021-05-12
[TEST] Migration of test_enable_storage_reserve_volume


aujjwal-redhat:
Migrate the [test_case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_enable_storage_reserve_volume.py).

Make the modifications as per the redant ops.

PR #318, opened 2021-05-12
Added test_enable_storage_reserve_volume

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_enable_storage_reserve_volume.py)

Modified as per the ops in redant.

Updates : #292
Fixes : #313

### DIFF
--- a/tests/functional/glusterd/test_enable_storage_reserve_volume.py
+++ b/tests/functional/glusterd/test_enable_storage_reserve_volume.py
@@ -1,0 +1,79 @@
+#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+"""
+Description:
+   This test case is authored to test posix storage.reserve option.
+"""
+
+from glusto.core import Glusto as g
+from glustolibs.gluster.exceptions import ExecutionError
+from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
+from glustolibs.gluster.volume_ops import set_volume_options
+
+
+@runs_on([['distributed-replicated'], ['glusterfs']])
+class TestPosixStorageReserveOption(GlusterBaseClass):
+    def setUp(self):
+        """
+        setUp method for every test.
+        """
+        self.get_super_method(self, 'setUp')()
+
+        # setup volume
+        ret = self.setup_volume()
+        if not ret:
+            raise ExecutionError("Failed to setup volume")
+
+    def test_enable_storage_reserve_volume(self):
+        """
+        1) Create a distributed-replicated volume and start it.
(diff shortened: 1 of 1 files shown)


tests/functional/glusterd/test_enable_storage_reserve_volume.py
```diff
@@ -0,0 +1,49 @@
...
+            let's say, set it to a value of 50.
+        3) check df -h output of the mount point and backend bricks.
+        """
+        # Set volume option storage.reserve 50
+        redant.set_volume_options(
+            self.vol_name, {"storage.reserve ": 50},
+            self.server_list[0])
+
+        # Get the mount point
+        mount_point = redant.get_mnt_pts_dict(self.vol_name)
```
nik-redhat, line 42: Does this function mount the volume as well?

tests/functional/glusterd/test_enable_storage_reserve_volume.py
```diff
@@ -0,0 +1,49 @@
...
+        mount_point = mount_point[self.client_list[0]][0]
+
+        # check df -h output
+        cmd = f"df -h | grep -i {mount_point}"
+        ret = redant.execute_command(cmd, self.client_list[0])
+        if "52%" not in ret['msg'][0].split(" "):
```
nik-redhat, line 48: The test case says 51%, then why change to 52%?

nik-redhat (changes requested):

tests/functional/glusterd/test_enable_storage_reserve_volume.py
aujjwal-redhat: The mounting operation takes place in environment setup itself. So, it got mounted before entering this test.

tests/functional/glusterd/test_enable_storage_reserve_volume.py
aujjwal-redhat: Yeah, actually I tested the command manually on all my servers and it was showing 52% instead. Moreover, I don't know why we are testing for the value. Since it is to check the output of the command itself, that means we should just check if output != [ ]

tests/functional/glusterd/test_enable_storage_reserve_volume.py
nik-redhat: But, what about tests which don't need to mount a volume....I am confident enough that there are a good number of tests which doesn't mount a volume. So, is this mounting limited to this particular test or in general?

srijan-sivakumar: If we want the mount operations to be handled inside the test case we can still do that. There's one category of TCs falling under Generic, in which none of these basic env setup is done (volume creation, start and mounting )

nik-redhat: Mounting is something which I believe shouldn't be done in general.
In fact, another way to reduce the time is to segregate the tests based on the task they do, like there are few tests that don't start a volume, and there by stopping is also not needed.

tests/functional/glusterd/test_enable_storage_reserve_volume.py
nik-redhat: Okay, so I checked the test steps and this can be modified a bit. The intention is to check that the set `50%` space should be visible, and we can simply modify it to check that the value returned should be more than equal to 50%.

aujjwal-redhat: Ok. Then let's change the condition to check if the value is >= 50. Is that ok @nik-redhat @srijan-sivakumar ?

nik-redhat: Yes.

aujjwal-redhat: Ack

nik-redhat:
You have tested the code, right?

aujjwal-redhat:
> You have tested the code, right?

yeah

nik-redhat (approved):

Issue #323, opened 2021-05-12
Migrate test_op_version


aujjwal-redhat:
Migrate the [testcase](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_op_version.py). Modify the test case as per the ops in redant.

PR #325, opened 2021-05-12
Added test_op_version

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_op_version.py).
Modified the same using the ops in redant

Updates: #292

Fixes: #323

### DIFF
--- a/tests/functional/glusterd/test_op_version.py
+++ b/tests/functional/glusterd/test_op_version.py
@@ -1,3 +1,4 @@
+"""
 #  Copyright (C) 2018  Red Hat, Inc. <http://www.redhat.com>
 #
 #  This program is free software; you can redistribute it and/or modify
@@ -14,48 +15,18 @@
 #  with this program; if not, write to the Free Software Foundation, Inc.,
 #  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
-""" Description:
+Description:
       Test for setting up the max supported op-version and
       verifying  version number in info file
 """
 
-from glusto.core import Glusto as g
-from glustolibs.gluster.exceptions import ExecutionError
-from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
-from glustolibs.gluster.glusterfile import file_exists
-from glustolibs.gluster.volume_ops import (get_volume_options,
-                                           set_volume_options)
+# nonDisruptive;dist-rep
+from tests.abstract_test import AbstractTest
 
 
-@runs_on([['distributed-replicated'], ['glusterfs']])
-class TestMaxSupportedOpVersion(GlusterBaseClass):
+class TestMaxSupportedOpVersion(AbstractTest):
 
-    def setUp(self):
-        """
-        setUp method for every test
-        """
-        # calling GlusterBaseClass setUp
-        self.get_super_method(self, 'setUp')()
-
-        # Creating Volume
-        g.log.info("Started creating volume")
-        ret = self.setup_volume()
-        if not ret:
-            raise ExecutionError("Volume creation failed: %s" % self.volname)
-
-    def tearDown(self):
-        """
-        tearDown for every test
-        """
-        # stopping the volume and Cleaning up the volume
-        ret = self.cleanup_volume()
-        if not ret:
-            raise ExecutionError("Failed Cleanup the Volume %s" % self.volname)
-
-        # Calling GlusterBaseClass tearDown
-        self.get_super_method(self, 'tearDown')()
-
-    def test_op_version(self):
+    def run_test(self, redant):
         '''
         -> Create Volume
         -> Get the current op-version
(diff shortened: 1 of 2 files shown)


aujjwal-redhat:
One thing that I noticed while converting this test case was :
In the following set of statements it mentions to grep the command and perform set volume operation
Screenshot from 2021-05-12 15-51-40
But it does not perform any volume set operation.

Moreover, in comments it is asking to set the volume and then compare the version but it's not setting the version. It is just updating the value of version and then leaving it as it is.

Screenshot from 2021-05-12 15-53-24

 In that case it was failing always as obviously there is no version update as no volume set operation took place.

Hence, added these lines in the migrated code
Screenshot from 2021-05-12 15-51-59


Let me know if it is not correct.

nik-redhat:
I think the code is fine.
The steps say `If current op-version is less than max-op-version set the current op-version to max-op-version`, which means it will compare the op-version and then update the op-version if not equal. And, then again the next step to compare the version is done only if the set operation is done.

tests/functional/glusterd/test_op_version.py
```diff
@@ -0,0 +1,97 @@
...
+        current_op_version = int(ret['cluster.op-version'])
+
+        # Getting Max op-verison
+        max_op_version = int(ret['cluster.max-op-version'])
+
+        # File_path: path for vol info file
+        # Checking vol file exist in all servers or not
+        file_path = '/var/lib/glusterd/vols/' + self.vol_name + '/info'
+        for server in self.server_list:
+            cmd = f'ls -ld {file_path}'
```
nik-redhat, line 54: The function that is used in the glusto test is good and should be implemented in the ops library. But, I don't know how are you guys tracking which function to be created.

aujjwal-redhat:
> I think the code is fine.
> The steps say `If current op-version is less than max-op-version set the current op-version to max-op-version`, which means it will compare the op-version and then update the op-version if not equal. And, then again the next step to compare the version is done only if the set operation is done.

But they are asking about version number. Not the cluster version number. If you notice there are two comments:
```bash
     -> After vol set operation verify that version number
        increased by one or not in vol info file
    -> verify that current-op-version and max-op-version same or not.
```
Why will they verify the same twice.

Moreover, they are gripping version in the prev statements and updating it by 1 but not setting the value in the version number. Then why to grep at first place and update it. And then the comment says to verify that version number.

tests/functional/glusterd/test_op_version.py
aujjwal-redhat: Thats the same function actually. Actually we can add this to the libs as it is in glusto. For now, I think so we can keep a todo.

aujjwal-redhat: This is the function in glusto:
```bash
def file_exists(host, filename):
    """Check if file exists at path on host

    Args:
        host (str): hostname or ip of system
        filename (str): fully qualified path of file

    Returns:
        True if file exists. False if file does not exist
    """
    command = "ls -ld %s" % filename
    rcode, _, _ = g.run(host, command)
    if rcode == 0:
        return True

    return False
```

nik-redhat: Aree....that I understand. But, what I meant was good to have a generic function which does this instead of writing the same piece of code again.
It can be tackled in a separate PR no issues with that, but my question was how do we remember which functions to add, do you all maintain any sort of doc or so?

aujjwal-redhat: No doc now. But just TODOs. :sweat_smile:

nik-redhat: Cool, so let's add a TODO on top of this code.

aujjwal-redhat: Ack.
I will modify the rest of the code as well. Thanks for the explanation :)

srijan-sivakumar: If it is easy to create that function now, lets do it now, rather than putting a TODO, gluster is an example of TODOs across the code base. Let's not repeat it here.

srijan-sivakumar: @aujjwal-redhat lets add the function now...I guess it won't be much of a hassle :p

aujjwal-redhat: ok. I will add this.

aujjwal-redhat:
@srijan-sivakumar I have added the function in IO ops.

tests/functional/glusterd/test_op_version.py
```diff
@@ -0,0 +1,94 @@
...
+        for server in self.server_list:
+            if not redant.file_exists(file_path, server):
+                raise Exception("File does not exist")
+
+        # Getting version number from vol info file
+        # cmd: grepping  version from vol info file
+        file_path_cmd = ' '.join(['grep', "'^version'", file_path])
+        ret = redant.execute_command(file_path_cmd, self.server_list[0])
+        out = ret['msg'][0]
+        version_list = out.split('=')
```
nik-redhat, line 62: This could be merged under a single operation.

tests/functional/glusterd/test_op_version.py
```diff
@@ -0,0 +1,94 @@
...
+            ret = redant.set_volume_options(
+                'all', {'cluster.op-version': max_op_version},
+                self.server_list[0])
+
+            # Grepping version number from vol info file after
+            # vol set operation
+            ret = redant.execute_command(file_path_cmd,
+                                         self.server_list[0])
+            out = ret['msg'][0]
+            version_list = out.split('=')
```
nik-redhat, line 78: Same as above

nik-redhat (changes requested):

tests/functional/glusterd/test_op_version.py
aujjwal-redhat: Ack

nik-redhat (approved):

Issue #315, opened 2021-05-12
[TEST] Migration of test_gluster_detect_drop_of_outbound_traffic


nishith-vihar:
Migrating the [test_gluster_detect_drop_of_outbound_traffic.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_gluster_detect_drop_of_outbound_traffic.py) test case

PR #327, opened 2021-05-12
Migration of test_gluster_detect_drop_of_outbound_traffic.py

nishith-vihar:
test case in glusto: [test_gluster_detect_drop_of_outbound_traffic.py](https://github.com/gluster/glusto-tests/blob/e1ea6ec5270dd89ec3b4d75c37398dc2e1cf3381/tests/functional/glusterd/test_gluster_detect_drop_of_outbound_traffic.py)

Updates: #292 
Fixes: #315 


tests/functional/glusterd/test_gluster_detect_drop_of_outbound_traffic.py
```diff
@@ -0,0 +1,83 @@
...
+        4) Execute few Gluster CLI commands like volume status, peer status
+        5) Gluster CLI commands should fail with suitable error message
+        """
+        server1 = self.server_list[0]
+
+        # Set iptablerule_set as false initially
+        iptablerule_set = False
+
+        # Set iptable rule on one node to drop outbound glusterd traffic
+        cmd = "iptables -I OUTPUT -p tcp --dport 24007 -j DROP"
```
srijan-sivakumar, line 44: @nishith-vihar if we are modifying ip table rules, this is a pure disruptive test case..I think that needs to be changes above in the configuration.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_gluster_detect_drop_of_outbound_traffic.py
nishith-vihar: Ack

nishith-vihar:
There is a part in the glusto-test where the stderr from the volume status command is checked to see if the number of nodes from the volume status command  is equal to the number of nodes from the peer list. But the volume status command is expected to fail in the test case and according to the current design in the ops, if the volume status command fails it raises exception and does not return anything. As of now that part is omitted in the current test migration. Any suggestions on how that can be done or can it be actually omitted?

aujjwal-redhat:
> There is a part in the glusto-test where the stderr from the volume status command is checked to see if the number of nodes from the volume status command is equal to the number of nodes from the peer list. But the volume status command is expected to fail in the test case and according to the current design in the ops, if the volume status command fails it raises exception and does not return anything. As of now that part is omitted in the current test migration. Any suggestions on how that can be done or can it be actually omitted?

@nishith-vihar try to raise the exception in such tests. 
For reference :
https://github.com/srijan-sivakumar/redant/blob/main/tests/functional/glusterd/test_change_reserve_limit_to_wrong_values.py

nishith-vihar:
> > There is a part in the glusto-test where the stderr from the volume status command is checked to see if the number of nodes from the volume status command is equal to the number of nodes from the peer list. But the volume status command is expected to fail in the test case and according to the current design in the ops, if the volume status command fails it raises exception and does not return anything. As of now that part is omitted in the current test migration. Any suggestions on how that can be done or can it be actually omitted?
> 
> @nishith-vihar try to raise the exception in such tests.
> For reference :
> https://github.com/srijan-sivakumar/redant/blob/main/tests/functional/glusterd/test_change_reserve_limit_to_wrong_values.py

@aujjwal-redhat I have raised the exception in the test case but we need the return value from volume_status to comapre the number of nodes returned by volume_status and peer list which is not received when exception is raised.

srijan-sivakumar:
> There is a part in the glusto-test where the stderr from the volume status command is checked to see if the number of nodes from the volume status command is equal to the number of nodes from the peer list. But the volume status command is expected to fail in the test case and according to the current design in the ops, if the volume status command fails it raises exception and does not return anything. As of now that part is omitted in the current test migration. Any suggestions on how that can be done or can it be actually omitted?

Straightforward way would require to add a different set of ops here. Now to be honest this is a rare test case. Not all test cases would require us to do so. But we need to think this through properly. We don't want to end up with a huge debt of functions used only once or maybe not even once.

The said situation requires that we let the command fail, but here's my doubt, firstly what do you mean by `volume_status` should fail, can you state how it is failed ?

nik-redhat:
Test run:
```
Summary:
Table:
 test_gluster_detect_drop_of_outbound_traffic
 
|     rep     |     PASS    | 61.40222525596619 |
|     dist    |     PASS    | 68.90594267845154 |
|     disp    |     PASS    | 67.90412378311157 |
|     arb     |     PASS    | 66.79957914352417 |
|   dist-arb  |     PASS    | 70.38102626800537 |
|   dist-rep  |     PASS    | 71.00289559364319 |


Framework runtime : 415.80915927886963
```

srijan-sivakumar (approved):

Issue #331, opened 2021-05-12
Gluster Ops : is_glusterd_running and wait_for_glusterd_to_start


aujjwal-redhat:
* In gluster ops:                                                               
 - wait_for_glusterd to start is explicitly converting the node to a           
    list even though is_glusterd_running is taking an str. This is leading to the dispatcher issue saying 
![Screenshot from 2021-05-12 21-22-31](https://user-images.githubusercontent.com/77244483/118005682-3ad5e880-b368-11eb-8ded-7e92004ad708.png)

   - wait_for_glusterd is returning false even when glusterd is running because the condition is wrong.        
 It should be `if ret` instead of `if not ret`. And because of that          
 in the logs 80 times the command ran successfully and still returned            
 false.

```bash
  def is_glusterd_running(self, node: str) -> bool:
        """
        Checks the status of the glusterd service on the
        specified node.
        Args:
            node (str): The node on which the glusterd service
                        is to be stopped.
        Returns:
            1  : if glusterd active
            0  : if glusterd not active
           -1  : if glusterd not active and PID is alive
        """
        is_active = 1
```

```bash
def wait_for_glusterd_to_start(.......)
.
.
 while count <= timeout:
            ret = self.is_glusterd_running(node)
            if not ret:
                return True
           sleep(1)
            count += 1
        return False
```
So, even when is_active is 1. if not 1 is false and it runs for 80 times.

PR #332, opened 2021-05-12
Fixed wait_for_glusterd issues

aujjwal-redhat:
* Removed the condition to explicitly convert node to list
* Changed the condition to return True if ret is 1.

Fixes: #331

### DIFF
--- a/common/ops/gluster_ops/gluster_ops.py
+++ b/common/ops/gluster_ops/gluster_ops.py
@@ -219,13 +219,10 @@ class GlusterOps(AbstractOps):
         Returns:
             bool: True if glusterd is running on the node(s) or else False.
         """
-        if node is None:
-            return False
-
         count = 0
         while count <= timeout:
             ret = self.is_glusterd_running(node)
-            if ret:
+            if ret == 1:
                 return True
             sleep(1)
             count += 1
--- a/common/ops/gluster_ops/gluster_ops.py
+++ b/common/ops/gluster_ops/gluster_ops.py
@@ -212,7 +212,7 @@ class GlusterOps(AbstractOps):
         it till the timeout is reached.
 
         Args:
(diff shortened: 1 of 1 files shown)


common/ops/gluster_ops/gluster_ops.py
```diff
@@ -212,22 +212,20 @@ def wait_for_glusterd_to_start(self, node=None, timeout: int = 80):
...
             timeout (int) : We cannot wait till eternity right :p
 
         Returns:
             bool: True if glusterd is running on the node(s) or else False.
         """
-        if not isinstance(node, list) and node is not None:
-            node = [node]
-        elif node is None:
+        if node is None:
             return False
```
nik-redhat, line 223: Doesn't make sense to check this. Otherwise, we would be checking this in all the OPs functions.

common/ops/gluster_ops/gluster_ops.py
```diff
@@ -212,22 +212,20 @@ def wait_for_glusterd_to_start(self, node=None, timeout: int = 80):
...
 
         count = 0
         while count <= timeout:
             ret = self.is_glusterd_running(node)
-            if not ret:
+            if ret:
```
nik-redhat, line 228: It should be `if ret==1`, else it will pass of `-1` too.

nik-redhat (changes requested):

common/ops/gluster_ops/gluster_ops.py
aujjwal-redhat: Ack

common/ops/gluster_ops/gluster_ops.py
aujjwal-redhat: Ack

nik-redhat (approved):

Issue #317, opened 2021-05-12
[TEST] Migration of test_glusterd_logs_when_peer_detach


nishith-vihar:
Migration of [test_glusterd_logs_when_peer_detach.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py) test case

PR #336, opened 2021-05-14
Migration of test_glusterd_logs_when_peer_detach

nishith-vihar:
[test_glusterd_logs_when_peer_detach.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py) is migrated to redant

Updates: #292
Fixes: #317  

### DIFF
--- a/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
+++ b/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
@@ -16,8 +16,7 @@
   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
   Description:
-        No Errors should generate in glusterd.log while detaching
-        node from gluster
+        Check for error in glusterd logs when a peer is detached
 """
 
 import random
@@ -48,7 +47,7 @@ class GlusterdLogsWhilePeerDetach(AbstractTest):
                                               self.server_list[0])
 
         # Clearing the existing glusterd log file
-        ret = redant.execute_abstract_op_node('echo > /var/log/glusterfs/'
+        ret = redant.execute_abstract_op_node('> /var/log/glusterfs/'
                                               'glusterd.log',
                                               self.server_list[0])
 
@@ -61,5 +60,5 @@ class GlusterdLogsWhilePeerDetach(AbstractTest):
                                               "glusterd.log | wc -l",
                                               self.server_list[0])
         if int(ret['msg'][0]) != 0:
-            redant.logger.info("Found Error messages in glusterd log "
-                               "file after peer detach")
+            raise Exception("Found Error messages in glusterd "
+                            "log file after peer detach")
--- a/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
+++ b/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
@@ -41,13 +41,16 @@ class GlusterdLogsWhilePeerDetach(AbstractTest):
         timestamp = ret['msg'][0].rstrip("\n")
 
         #  glusterd logs
-        ret = redant.execute_abstract_op_node('cp /var/log/glusterfs/glusterd.log /var/'
-                                    f'log/glusterfs/glusterd_{timestamp}.log',
-                                    self.server_list[0])
+        ret = redant.execute_abstract_op_node('cp /var/log/glusterfs/'
+                                              'glusterd.log /var/log/'
+                                              'glusterfs/'
+                                              f'glusterd_{timestamp}.log',
+                                              self.server_list[0])
 
         # Clearing the existing glusterd log file
-        ret = redant.execute_abstract_op_node('echo > /var/log/glusterfs/glusterd.log',
-                                    self.server_list[0])
+        ret = redant.execute_abstract_op_node('echo > /var/log/glusterfs/'
+                                              'glusterd.log',
+                                              self.server_list[0])
 
         # Performing peer detach
         random_server = random.choice(self.server_list[1:])
@@ -55,8 +58,8 @@ class GlusterdLogsWhilePeerDetach(AbstractTest):
 
         # Searching for error message in log
         ret = redant.execute_abstract_op_node("grep ' E ' /var/log/glusterfs/"
-                                    "glusterd.log | wc -l",
-                                    self.server_list[0])
+                                              "glusterd.log | wc -l",
+                                              self.server_list[0])
         if int(ret['msg'][0]) != 0:
             redant.logger.info("Found Error messages in glusterd log "
                                "file after peer detach")
--- a/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
+++ b/tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
@@ -1,52 +1,34 @@
-#  Copyright (C) 2018-2020  Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License along
-#  with this program; if not, write to the Free Software Foundation, Inc.,
-#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+  Copyright (C) 2018-2020  Red Hat, Inc. <http://www.redhat.com>
 
-""" Description:
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License along
+  with this program; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+  Description:
         No Errors should generate in glusterd.log while detaching
         node from gluster
 """
 
 import random
-from glusto.core import Glusto as g
-from glustolibs.gluster.exceptions import ExecutionError
-from glustolibs.gluster.gluster_base_class import GlusterBaseClass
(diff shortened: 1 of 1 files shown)


tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,62 @@
...
+    def run_test(self, redant):
+        '''
+        -> Detach the node from peer
+        -> Check that any error messages related to peer detach
+        in glusterd log file
+        -> No errors should be there in glusterd log file
+        '''
+
+        # Getting timestamp
+        ret = redant.execute_io_cmd('date +%s', self.server_list[0])
```
nik-redhat, line 40: Why this function and not `execute_command`?

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # Performing peer detach
+        random_server = random.choice(self.server_list[1:])
+        ret = redant.peer_detach(random_server, self.server_list[0])
+
+        # Searching for error message in log
+        ret = redant.execute_io_cmd("grep ' E ' /var/log/glusterfs/"
+                                    "glusterd.log | wc -l",
+                                    self.server_list[0])
+        if int(ret['msg'][0].rstrip("\n")) != 0:
```
nik-redhat, line 60: No need for rstrip here, and I don't think we need it above as well. Could you try out without rstrip and see if that causes an issue?

nik-redhat (changes requested):

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: the timestamp here is used in creation of the logfile. without rstrip there will be a '\n' at the end of timetamp

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: In the above case the timestamp is used in creating a log file. without rstrip it will be left with '\n' at the end.

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,62 @@
...
+# disruptive;
+
+
+class GlusterdLogsWhilePeerDetach(AbstractTest):
+
+    def run_test(self, redant):
+        '''
+        -> Detach the node from peer
+        -> Check that any error messages related to peer detach
+        in glusterd log file
```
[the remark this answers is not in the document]
nishith-vihar: Ack

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: This func will be replaced by execute_abstract_op_node function.

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nik-redhat: Have you tried both cases?

nishith-vihar: I have tried it now. in the case of timestamp it is required because it is giving out error due to the \n and log file creation is not proper but in the second case in is not necessary.

nik-redhat: Okay.

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,65 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License along
+  with this program; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+  Description:
+        No Errors should generate in glusterd.log while detaching
+        node from gluster
```
nik-redhat, line 20: Let's make it:
`Check for error in glusterd logs when a peer is detached`

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,65 @@
...
+
+        #  glusterd logs
+        ret = redant.execute_abstract_op_node('cp /var/log/glusterfs/'
+                                              'glusterd.log /var/log/'
+                                              'glusterfs/'
+                                              f'glusterd_{timestamp}.log',
+                                              self.server_list[0])
+
+        # Clearing the existing glusterd log file
+        ret = redant.execute_abstract_op_node('echo > /var/log/glusterfs/'
```
srijan-sivakumar, line 51: nit: instead of `echo > /var/log/glusterfs/glusterd.log`, we can directly use `> /var/log/glusterfs/glusterd.log`

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,65 @@
...
+        # Performing peer detach
+        random_server = random.choice(self.server_list[1:])
+        ret = redant.peer_detach(random_server, self.server_list[0])
+
+        # Searching for error message in log
+        ret = redant.execute_abstract_op_node("grep ' E ' /var/log/glusterfs/"
+                                              "glusterd.log | wc -l",
+                                              self.server_list[0])
+        if int(ret['msg'][0]) != 0:
+            redant.logger.info("Found Error messages in glusterd log "
```
srijan-sivakumar, line 64: We should raise an exception here rather than using the `redant.logger.error`, we should be doing something like
```        if int(ret['msg'][0]) != 0:
                 raise Exception("Found Error messages in glusterd log file after peer detach")```

**Review by srijan-sivakumar — CHANGES_REQUESTED** [review]

File: tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py (hunk 4 above)
Comment (nishith-vihar) [reply tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py:20]: Ack

File: tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py (hunk 5 above)
Comment (nishith-vihar) [reply tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py:51]: Ack

File: tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py (hunk 6 above)
Comment (nishith-vihar) [reply tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py]: Ack

**Comment by nishith-vihar** [comment]
Screenshot from 2021-05-14 15-11-02

File: tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py [hunk 7]
```diff
@@ -0,0 +1,64 @@
...
+                in glusterd log file
+            3) No errors should be there in glusterd log file
+        '''
+
+        # Getting timestamp
+        ret = redant.execute_abstract_op_node('date +%s', self.server_list[0])
+        timestamp = ret['msg'][0].rstrip("\n")
+
+        #  glusterd logs
+        ret = redant.execute_abstract_op_node('cp /var/log/glusterfs/'
```
nik-redhat, line 43: No need for the return value

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,64 @@
...
+                                              'glusterd.log /var/log/'
+                                              'glusterfs/'
+                                              f'glusterd_{timestamp}.log',
+                                              self.server_list[0])
+
+        # Clearing the existing glusterd log file
+        ret = redant.execute_abstract_op_node('> /var/log/glusterfs/'
```
nik-redhat, line 50: Same as above

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,64 @@
...
+        random_server = random.choice(self.server_list[1:])
+        ret = redant.peer_detach(random_server, self.server_list[0])
+
+        # Searching for error message in log
+        ret = redant.execute_abstract_op_node("grep ' E ' /var/log/glusterfs/"
+                                              "glusterd.log | wc -l",
+                                              self.server_list[0])
+        if int(ret['msg'][0]) != 0:
+            raise Exception("Found Error messages in glusterd "
+                            "log file after peer detach")
```
nik-redhat, line 64: This can be updated as well, to not check for ret value

nik-redhat (changes requested):

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: For raising the exception we need to check for the ret value right?

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: Ack

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: Ack

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nik-redhat: This check is done inside the `execute_abstract_op_node` function as well, so the exception will be raised from there.

nishith-vihar: no. actually there are no checks for ret['msg'] returned. It handles only the error_code exception.

nik-redhat: Oh my bad. It's needed in such a case.

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
```diff
@@ -0,0 +1,64 @@
...
+                                              self.server_list[0])
+
+        # Clearing the existing glusterd log file
+        ret = redant.execute_abstract_op_node('> /var/log/glusterfs/'
+                                              'glusterd.log',
+                                              self.server_list[0])
+
+        # Performing peer detach
+        random_server = random.choice(self.server_list[1:])
+        ret = redant.peer_detach(random_server, self.server_list[0])
```
nik-redhat, line 56: nit: ret not needed.

nishith-vihar: Ack

tests/functional/glusterd/test_glusterd_logs_when_peer_detach.py
nishith-vihar: Ack

nik-redhat (approved):

PR #349, opened 2021-05-14
Migration of test_validate_glusterd_info

nishith-vihar:
[test_validate_glusterd_info.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_validate_glusterd_info.py) test case is migrated to redant


tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,97 @@
...
+                raise Exception(f"UUID does not match in host{server}")
+
+            # Geting the UUID from cmd "gluster peer status"
+            for node in self.server_list:
+                peer_status = redant.get_peer_status(node)
+                if isinstance(peer_status, list):
+                    for i in redant.get_peer_status(node):
+                        uuid_list.append(i["uuid"])
+                else:
+                    uuid_list.append(peer_status["uuid"])
```
nik-redhat, line 67: I don't think the logic is correct here.
It should be like:
```
for node in self.server_list:
   uuid_list = []
   if node!=server:
       peer_status_list = redant.get_peer_status(node)
       if isinstance(peer_status_list, list):
           for peer in peer_status_list:
               uuid_list.append(i["uuid"])
       else
           uuid_list.append(peer_status_list["uuid"])

       if get_uuid.rstrip("\n") not in uuid_list:
           raise Exception(f"uuid not matched in {node}")
```

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,97 @@
...
+            raise Exception("Failed to get expected output")
+
+        # Checking for glusterd.service is enabled by default
+        ret = redant.execute_abstract_op_node(
+            "systemctl is-enabled glusterd.service", self.server_list[0])
+        msg = " ".join(ret['msg'])
+        if msg.replace("\n", "") != 'enabled':
+            raise Exception(
+                "Output of systemctl is-enabled glusterd.service is"
+                " not enabled")
```
nik-redhat, line 97: Better to define this function before `run_test`, also make this function private.

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,97 @@
...
+        ret = redant.execute_abstract_op_node(cmd, self.server_list[0])
+        msg = " ".join(ret['msg'])
+
+        # Checking glusterd.socket file
+        if msg.replace("\n", "") != "glusterd.socket":
+            raise Exception("Failed to get expected output")
+
+        # Checking for glusterd.service is enabled by default
+        ret = redant.execute_abstract_op_node(
+            "systemctl is-enabled glusterd.service", self.server_list[0])
```
nik-redhat, line 92: nit: indenetation

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,97 @@
...
+                else:
+                    uuid_list.append(peer_status["uuid"])
+
+                if server != node and get_uuid.rstrip("\n") not in uuid_list:
+                    raise Exception(f"uuid not matched in {node}")
+        self.test_glusterd_config_file_check(redant)
+
+    def test_glusterd_config_file_check(self, redant):
+        """
+        Steps:
```
nik-redhat, line 75: Steps should only be mentioned in the main function, i.e, the one which would be invoked directly, the other method's docstring should only say what that function does.

nik-redhat (changes requested):

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: Actually in glusto this is a complete different test function which is not related to the run_test function. I think py-test has the mechanism to run both the test funciton at run time but as we dont have that in redant i called the second test function fron the run_test function. That is why there are steps for both the functions

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: Ack

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,97 @@
...
+                "grep -i uuid /var/lib/glusterd/glusterd.info", server)
+            glusterd_volinfo = " ".join(ret['msg'])
+            uuid_list.append(glusterd_volinfo)
+            glusterd_volinfo = (glusterd_volinfo.split("="))[1]
+            if not glusterd_volinfo:
+                raise Exception("UUID not found in 'glusterd.info' file ")
+
+            # Getting UUID from cmd 'gluster system uuid get'
+            ret = redant.execute_abstract_op_node(
+                "gluster system uuid get | awk {'print $2'}", server)
```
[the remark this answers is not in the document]
nishith-vihar: Ack

tests/functional/glusterd/test_validate_glusterd_info.py
srijan-sivakumar: In this scenario, we should be seeing as to how to treat it as a different test case.

srijan-sivakumar: These two are separate test cases ( sub cases ). Instead of failing both the sub-tests because of failure in one of the subcases, it would be better to handle them separately.

nik-redhat: I don't think these are 2 different test cases or functions. This function is called in the last line of the previous function. For more verification, you can try to run these test on the shared setup and see what happens.
Then, comment out the last line where it calls the function, and see if the other function is still invoked or not ( I think it won't be).

nik-redhat: Ohh, now I see that the calling of the other function is added by you but it's not in the actual test case.

nishith-vihar: Actually the last line which is calling the function is added by me because i thought that in this manner we can execute both the test functions but if you look into the glusto-test it does not call the second function in the test.

srijan-sivakumar: @nik-redhat if one were to look at the original test case in glusto-tests, it does seem like two different sub-tests.

srijan-sivakumar: What we can do is to merge these two in one single test function. As mentioned before, we should also see if we can fuse test cases together rather than splitting them ( which we still need to do, if need be ).

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: Yeah. I think this logic makes sense. The logic i wrote was based on the glusto-test but maybe there was a mistake in the glusto-test case itself.

tests/functional/glusterd/test_validate_glusterd_info.py
nik-redhat: Okay, so on checking back the timeline it looks like there isn't any special purpose for doing so. These are actually 2 separate test cases with 2 different ID's can be separated out without any trouble.

tests/functional/glusterd/test_validate_glusterd_info.py
nik-redhat: Also, please add an SS of the successful run of the test case in each PR, and after every commit. Currently, we don't have a CI test so this is the only way to verify it.

nishith-vihar: Ack

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: In that case I will create another test case with the name of the function.

nishith-vihar:
Screenshot from 2021-05-17 12-49-45

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,71 @@
...
+            if glusterd_volinfo != get_uuid:
+                raise Exception(f"UUID does not match in host{server}")
+
+            # Geting the UUID from cmd "gluster peer status"
+            for node in self.server_list:
+                uuid_list = []
+                if node != server:
+                    peer_status = redant.get_peer_status(node)
+                    if isinstance(peer_status, list):
+                        for i in redant.get_peer_status(node):
```
nik-redhat, line 65: Why are calling the get_peer_status function again?
Simply iterating over the peer_status list should do, right?

nishith-vihar: Oh yeah. My bad. I will correct that

nishith-vihar:
Screenshot from 2021-05-17 14-00-29

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,94 @@
...
+            if glusterd_volinfo != get_uuid:
+                raise Exception(f"UUID does not match in host{server}")
+
+            # Geting the UUID from cmd "gluster peer status"
+            for node in self.server_list:
+                uuid_list = []
+                if node != server:
+                    peer_status = redant.get_peer_status(node)
+                    if isinstance(peer_status, list):
+                        for i in peer_status:
```
nik-redhat, line 70: nit: let's use meaningful variable names wherever possible. Instead of "i" we can use "peer", over here.

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,94 @@
...
+                    if isinstance(peer_status_list, list):
+                        for peer in peer_status_list:
+                            uuid_list.append(peer["uuid"])
+                    else:
+                        uuid_list.append(peer_status_list["uuid"])
+
+                    if get_uuid.rstrip("\n") not in uuid_list:
+                        raise Exception(f"uuid not matched in {node}")
+
+        # Merged test_glusterd_config_file_check into this test case
```
srijan-sivakumar, line 78: Can we replace this with a function in io_ops which is responsible for checking if a certain file exists ?

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: Actually the command which is formed here uses grep -i which ignores case but the current existing path_exists function in io_ops takes in the exact filename with path.

nishith-vihar: or maybe we can directly execute the command `ls /var/run/ | grep -i glusterd.socket` using execute_abstract_op_node command which also takes care of raising the exception if file is not found. how about that?

nik-redhat: I think the current approach is fine. If we don't get the desired file we would anyway be logging an error later on.

tests/functional/glusterd/test_validate_glusterd_info.py
```diff
@@ -0,0 +1,94 @@
...
+  You should have received a copy of the GNU General Public License along
+  with this program; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+
+# nonDisruptive;
+from tests.abstract_test import AbstractTest
+
+
+class TestGlusterdInfo(AbstractTest):
```
nik-redhat, line 23: It needs to be updated based on the latest update on the master branch for the test execution patterns.

nishith-vihar: I rebased it and made change

tests/functional/glusterd/test_validate_glusterd_info.py
nishith-vihar: Ack

nishith-vihar:
Screenshot from 2021-05-18 18-14-45

tests/functional/glusterd/test_validate_glusterd_info.py
srijan-sivakumar: The issue is by deductive logic, then most of the commands can be invoked in this manner. My concern is that we don't set a precedent that we can invoke all commands directly. I do agree that this won't be much of a issue, but let's create a function so that people don't have to write these lines again and again.

srijan-sivakumar (changes requested):

srijan-sivakumar (approved):
LGTM

nik-redhat (approved):

Issue #355, opened 2021-05-18
Migration of test_validate_peer_probe_ip_fqdn_hostname


nishith-vihar:
Migration of [test_validate_peer_probe_ip_fqdn_hostname.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_validate_peer_probe_ip_fqdn_hostname.py) test case from glusto to redant

PR #358, opened 2021-05-18
Migration of test_validate_peer_probe_ip_fqdn_hostname

nishith-vihar:
Migration of [test_validate_peer_probe_ip_fqdn_hostname.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_validate_peer_probe_ip_fqdn_hostname.py) test case

Updates: #292
Fixes: #355 


nishith-vihar:
The commented function `_verify_cmd_history` checks the command history for successful peer probe by checking the start and end of a test case in the log file but in out case we the starting and ending of a test case is not logged in the `/var/log/glusterfs/cmd_history.log` file. Is there a way in which this can be implemented or checking the command history in itself is redundant?

srijan-sivakumar:
> The commented function `_verify_cmd_history` checks the command history for successful peer probe by checking the start and end of a test case in the log file but in out case we the starting and ending of a test case is not logged in the `/var/log/glusterfs/cmd_history.log` file. Is there a way in which this can be implemented or checking the command history in itself is redundant?

Maybe instead of injecting, as this is a disruptive test case, we can clear out the cmd_history and then start the operations, hence we can be sure that the required log came because of our test case.

nishith-vihar:
Screenshot from 2021-05-19 14-48-07

srijan-sivakumar (approved):
LGTM

nik-redhat (approved):

Issue #364, opened 2021-05-18
Automated migration script in tool


srijan-sivakumar:
We need an automated migration script which will do the following,
1. Take in the path of the redant repository and the absolute path of the file inside the glusto-tests, 
2. Perform the migration to redant filepath.
3. The final feature is to have the parsing and modifications into the newly copied file so as to make it more redantish from glustoish.

PR #372, opened 2021-05-18
[Tools] Migration Script

srijan-sivakumar:
The migration script is now able to handle the
operation of automatically doing the migration of a
test case from glusto to redant with history.

What isn't done is the modifications in the said test file.
This change, i.e. automated script changes will be part
of subsequent PRs.

Updates: #364

### DIFF
--- a/tools/migrate_tc.py
+++ b/tools/migrate_tc.py
@@ -1,4 +1,7 @@
-""" Migration script for glusto to redant. To be used only for test cases."""
+""" 
+Migration script for test case migration from Glusto-tests to
+redant repository. For more information, check the help flag.
+"""
 
 from os import path, system, chdir
 import argparse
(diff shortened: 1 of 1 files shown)


srijan-sivakumar:
The next stop is the migration script which converts the transffered file into a mature redant file ( as much as possible )

tools/migrate_tc.py
```diff
@@ -0,0 +1,74 @@
+""" Migration script for glusto to redant. To be used only for test cases."""
```
nik-redhat, line 1: Adding a comment to highlight the usage of the script would be helpful.

srijan-sivakumar: Added a line for checking `help`. There are only two flags so, I it'd be easy for the user to just query it. Let's not clutter with describing something again and again as the usage is derived from the help itself.

aujjwal-redhat (approved):
LGTM

nik-redhat (approved):

Issue #377, opened 2021-05-19
get_volume_status to be handled in case volumes are None


nishith-vihar:
when the volumes are none then the get_volume_status function in the volume_ops should return None

PR #379, opened 2021-05-19
The case where volumes are none is handled.

nishith-vihar:
When volumes are none get_volume status function should
return None

Fixes: #377


nik-redhat (approved):

srijan-sivakumar (approved):
LGTM

Issue #387, opened 2021-05-19
Migrate Test_profile_start_with_quorum_not_met.py


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_profile_start_with_quorum_not_met.py)

PR #397, opened 2021-05-20
Migrated test_profile_start_with_quorum_not_met

aujjwal-redhat:
Migrated the [test case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_profile_start_with_quorum_not_met.py).

Added validate_peer_is_connected in peer ops. It's originally in Gluster
Base class in glusto-tests.

Updates: #292
Fixes: #387

### DIFF
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -277,3 +277,38 @@ class PeerOps(AbstractOps):
             sleep(1)
             count += 1
         return False
+
+    def validate_peers_are_connected(self, server_list: list,
+                                     node: str) -> bool:
+        """
+        Validate whether each server in the cluster is connected to
+        all other servers in cluster.
+
+        Returns (bool): True if all peers are in connected
+                        state with other peers.
+                        False otherwise.
+        """
+        # if the setup has single node cluster
+        # then by-pass this validation
+        if len(server_list) == 1:
+            return True
+
+        # validate if peer is connected from all
+        # the servers
+        self.logger.info(f"Validating if {server_list} are connnected in"
+                         f" the cluster")
+
+        for server in server_list:
+            self.logger.info(f"Is peer connected {server}->{server_list}")
+            ret = self.is_peer_connected(server, server_list)
+            self.logger.info(f"Peer {server} is connected")
+
+            if not ret:
+                self.logger.error(f"Servers are not in connected state"
+                                  f" from {server}")
+                return False
+
+        self.logger.info("Successfully validated. All the servers"
+                         " in the cluster are connected")
+        self.get_peer_status(node)
+        return True
--- a/tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
+++ b/tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
@@ -1,58 +1,36 @@
-#  Copyright (C) 2017-2019  Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License along
-#  with this program; if not, write to the Free Software Foundation, Inc.,
-#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+Copyright (C) 2017-2019  Red Hat, Inc. <http://www.redhat.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+Description:
+
+This test case deals with starting profile operations when quorum is
+not met.
+"""
+
+# disruptive;dist,rep,disp,dist-rep,dist-disp
 
 from random import choice
 from time import sleep
-from glusto.core import Glusto as g
-from glustolibs.gluster.exceptions import ExecutionError
-from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
-from glustolibs.gluster.volume_ops import set_volume_options
-from glustolibs.gluster.gluster_init import start_glusterd, stop_glusterd
-from glustolibs.gluster.profile_ops import profile_start, profile_stop
+from tests.d_parent_test import DParentTest
 
 
-@runs_on([['distributed', 'replicated', 'dispersed',
-           'distributed-replicated', 'distributed-dispersed'], ['glusterfs']])
-class TestProfileStartWithQuorumNotMet(GlusterBaseClass):
+class TestCase(DParentTest):
 
-    def setUp(self):
-
-        # calling GlusterBaseClass setUp
-        self.get_super_method(self, 'setUp')()
-
-        # Creating Volume
-        g.log.info("Started creating volume")
-        ret = self.setup_volume()
-        if not ret:
-            raise ExecutionError("Volume creation failed: %s" % self.volname)
-        g.log.info("Volme created successfully : %s", self.volname)
-
-    def tearDown(self):
-
-        # stopping the volume and Cleaning up the volume
-        ret = self.cleanup_volume()
-        if not ret:
-            raise ExecutionError("Failed Cleanup the Volume %s" % self.volname)
-        g.log.info("Volume deleted successfully : %s", self.volname)
-
-        # Calling GlusterBaseClass tearDown
-        self.get_super_method(self, 'tearDown')()
-
-    def test_profile_start_with_quorum_not_met(self):
-        # pylint: disable=too-many-statements
+    def run_test(self, redant):
         """
         1. Create a volume
         2. Set the quorum type to server and ratio to 90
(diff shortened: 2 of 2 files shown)


aujjwal-redhat:
Screenshot from 2021-05-20 15-02-27
This is the one where it is getting timed out in dist-disp case.

nik-redhat:
Does that test fail always or occasionally?

aujjwal-redhat:
> Does that test fail always or occasionally?

Dist-disp fails most of the times unless I reduce the count in validation.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,95 @@
...
+
+        try:
+            redant.profile_start(self.vol_name, choice(new_servers))
+
+        except Exception:
+            redant.logger.info("Profile start failed as expected")
+
+        # Start glusterd on the node where it is stopped
+        redant.start_glusterd(node_on_glusterd_to_stop)
+        redant.logger.info(f"Successfully started glusterd "
```
srijan-sivakumar, line 81: Not required. As we would be already logging that we are going to start glusterd. It was thought not to have a separate success log.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # Setting Quorum ratio to 90%
+        quorum_perecent = {'cluster.server-quorum-ratio': '90%'}
+        redant.set_volume_options('all', quorum_perecent,
+                                  self.server_list[0])
+
+        # Stop glusterd on one of the node randomly
+        node_on_glusterd_to_stop = choice(self.server_list[1:])
+        redant.stop_glusterd(node_on_glusterd_to_stop)
+        redant.logger.info(f"glusterd stop on the node: "
```
srijan-sivakumar, line 57: Same..we can remove this success log.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
nik-redhat: Do we log this in the ops? If so then it's fine, but if not then this can be kept I feel.

nik-redhat: But, if we don't want it totally then I guess the same needs to be updated for other scenarios in other TCs as well.

common/ops/gluster_ops/peer_ops.py
```diff
@@ -277,3 +277,38 @@ def wait_for_peers_to_connect(self, node: str, servers: list,
             sleep(1)
             count += 1
         return False
+
+    def validate_peers_are_connected(self, server_list: list,
+                                     node: str) -> bool:
```
nik-redhat, line 282: If you are passing the node and servers on which the command is run and the servers to be checked respectively, how come this function is different from `is_peer_connected`?
Check the implementation in glusto it's a bit different. It checks runs the command in all the servers present in the cluster and check if all the peers are connected to each other, not a subset.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,91 @@
...
+        node_on_glusterd_to_stop = choice(self.server_list[1:])
+        redant.stop_glusterd(node_on_glusterd_to_stop)
+
+        # checking whether peers are connected or not
+        for _ in range(5):
+            ret = redant.validate_peers_are_connected(self.server_list[:],
+                                                      self.server_list[0])
+            if ret:
+                redant.logger.error("Peers are in connected state even after "
+                                    "stopping glusterd on one node")
```
nik-redhat, line 64: This check should be outside the loop. And instead, we should break if we get false from the `validate_peers_are_connected()`. Check the logic in the glusto TC.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,91 @@
...
+            sleep(2)
+
+        # Starting volume profile when quorum is not met
+        new_servers = self.server_list[:]
```
nik-redhat, line 68: Is `self.server_list` != `self.server_list[:]`?

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Start glusterd on the node where it is stopped
+        redant.start_glusterd(node_on_glusterd_to_stop)
+
+        for _ in range(5):
+            ret = redant.validate_peers_are_connected(self.server_list[:],
+                                                      self.server_list[0])
+            if not ret:
+                redant.logger.error("Peers are not in connected state")
+            sleep(2)
```
nik-redhat, line 85: Re-check the logic. Same as above.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Starting profile when volume quorum is met
+        redant.profile_start(self.vol_name, self.server_list[0])
+
+        # Stop the profile
+        redant.profile_stop(self.vol_name, self.server_list[0])
```
nik-redhat, line 91: @srijan-sivakumar handling of stopping profile in case it has started and the TC failed before stopping it needs to be done via ops and environ file. Or is it the case that profile TCs would always be `disruptive`?

nik-redhat (changes requested):

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
srijan-sivakumar: Yes, this needs to be checked. But..there's a catch. Not all cases be working on profile so it makes sense to let it be stopped in the TC and maybe if required have a custom teardown just for this. But if the test case is failing, I feel it is genuine. Also, would it affect the volume stop and delete if the profile isn't stopped ?

nik-redhat: Yes, it will be stopped in the TC, but suppose if the `profile stop` command failed, then the `profile` would still be running on all/the respective volumes. And, even if we add a teardown it might or might not help, depends upon the type of failure, because if we would have deleted the volume in the teardown then it's of no worry but here we would try to do a second `profile_stop` which might fail again and it will not help.
The issue with the profile not been stopped if it's not used is memory consumption. It consumes a healthy amount of resources if started for a long period of time.

common/ops/gluster_ops/peer_ops.py
aujjwal-redhat: No node is not where the command is run. Node is just for peer status. Even glusto is doing the same.

aujjwal-redhat: I have used the same logic as of glusto
```js
    @classmethod
    def validate_peers_are_connected(cls):
        """Validate whether each server in the cluster is connected to
        all other servers in cluster.

        Returns (bool): True if all peers are in connected with other peers.
            False otherwise.
        """

        # If the setup has single node server, by pass this validation.
        if len(cls.servers) == 1:
            return True

        # Validate if peer is connected from all the servers
        g.log.info("Validating if servers %s are connected from other servers "
                   "in the cluster", cls.servers)
        for server in cls.servers:
            g.log.info("Validate servers %s are in connected from  node %s",
                       cls.servers, server)
            ret = is_peer_connected(server, cls.servers)
            if not ret:
                g.log.error("Some or all servers %s are not in connected "
                            "state from node %s", cls.servers, server)
                return False
            g.log.info("Successfully validated servers %s are all in "
                       "connected state from node %s",
                       cls.servers, server)
        g.log.info("Successfully validated all servers %s are in connected "
                   "state from other servers in the cluster", cls.servers)

        # Peer Status from mnode
        peer_status(cls.mnode)
```

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
aujjwal-redhat: Ack.

common/ops/gluster_ops/peer_ops.py
nik-redhat: That's because you are passing the same args which the implementation of glusto uses. But, if someone wants to pass different args he will be able to do so, and that makes this function work exactly same as `is_peer_connected`

nik-redhat: The catch here is to not pass any arguments and decide the working in the function itself, that is execute the command for all the nodes just as in the glusto implementation.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
aujjwal-redhat: > self.server_list != self.server_list[:]
By value yes they are same
By reference no

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
nik-redhat: Ohh, gotcha. Python tricks :+1:

common/ops/gluster_ops/peer_ops.py
aujjwal-redhat: Glusto is using this function in gluster base class and hence it is able to use the servers as you can see here.
```js
for server in cls.servers:
            g.log.info("Validate servers %s are in connected from  node %s",
                       cls.servers, server)
            ret = is_peer_connected(server, cls.servers)

```
But in our implementation it is in peer ops. We will need to provide the server_list in that case and even the node to get the status. If this function is added to a place like lazy parent test, then yeah, we won't need any args to pass but since it is in the ops, we need to send the server_list and the node.

nik-redhat: Ack.

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # Stop glusterd on one of the node randomly
+        node_on_glusterd_to_stop = choice(self.server_list[1:])
+        redant.stop_glusterd(node_on_glusterd_to_stop)
+
+        # checking whether peers are connected or not
+        for _ in range(5):
+            ret = redant.validate_peers_are_connected(self.server_list[:],
+                                                      self.server_list[0])
+            if ret:
```
nik-redhat, line 62: It should be `if not ret`. It might return `true` within the loop, that's why we actually have the loop, to run it for 10secs and then check if we still have the same result or not. If we fail in the first iteration itself then what's the use of having a loop?

tests/functional/glusterd/test_profile_start_with_quorum_not_met.py
```diff
@@ -0,0 +1,97 @@
...
+        except Exception:
+            redant.logger.info("Profile start failed as expected")
+
+        # Start glusterd on the node where it is stopped
+        redant.start_glusterd(node_on_glusterd_to_stop)
+
+        for _ in range(5):
+            ret = redant.validate_peers_are_connected(self.server_list[:],
+                                                      self.server_list[0])
+            if not ret:
```
nik-redhat, line 86: It should be `if ret`

aujjwal-redhat: Ack

aujjwal-redhat:
```bash
redantenv) [root@servera redant]# python3 core/redant_main.py -c ../Utilities/conf.yaml -ll D -t tests/functional/glusterd/test_profile_start_with_quorum_not_met.py -sp
test_profile_start_with_quorum_not_met-dist PASS

test_profile_start_with_quorum_not_met-rep PASS

test_profile_start_with_quorum_not_met-disp PASS

test_profile_start_with_quorum_not_met-dist-rep PASS

test_profile_start_with_quorum_not_met-dist-disp FAIL

Table:
 test_profile_start_with_quorum_not_met
 
|     dist    |     PASS    | 36.38031530380249  |
|     rep     |     PASS    | 26.74582290649414  |
|     disp    |     PASS    | 63.19200396537781  |
|   dist-rep  |     PASS    | 62.18240308761597  |
|  dist-disp  |     FAIL    | 146.96898293495178 |

Framework runtime : 340.3483729362488
```

nik-redhat (approved):
LGTM

srijan-sivakumar (approved):

Issue #403, opened 2021-05-20
[Test] Migrating test_readonly_option_on_volume


srijan-sivakumar:
To migrate the test case : [test_readonly_option_on_volume](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_readonly_option_on_volume.py)

PR #404, opened 2021-05-20
[Test] Migrating test_readonly_option_on_volume

srijan-sivakumar:
Test migration :  [test_readonly_option_on_volume](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_readonly_option_on_volume.py)

Fixes: #403

### DIFF
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -27,7 +27,8 @@ class IoOps(AbstractOps):
         cmd = f"touch {path}/{filename}"
         self.execute_abstract_op_node(cmd, node)
 
-    def create_dir(self, path: str, dirname: str, node: str, excep: bool=True):
+    def create_dir(self, path: str, dirname: str, node: str,
+                   excep: bool = True):
         """
         Creates a directory in the specified path
         Args:
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -27,7 +27,7 @@ class IoOps(AbstractOps):
         cmd = f"touch {path}/{filename}"
         self.execute_abstract_op_node(cmd, node)
 
-    def create_dir(self, path: str, dirname: str, node: str):
+    def create_dir(self, path: str, dirname: str, node: str, excep: bool=True):
         """
         Creates a directory in the specified path
         Args:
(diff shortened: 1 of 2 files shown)


tests/functional/glusterd/test_readonly_option_on_volume.py
```diff
@@ -0,0 +1,62 @@
...
+# nonDisruptive;dist,rep,dist-rep,disp,dist-disp
+
+
+class TestCase(NdParentTest):
+
+    def run_test(self, redant):
+        """
+        Test steps:
+        1. Create a volume
+        2. Mount the volume
```
nishith-vihar, line 34: I think these steps can be removed as a part of the core test case steps

srijan-sivakumar: Let these steps be there as it'd help a layman or someone just writing a test case as to what this test case is doing and what are the steps.

tests/functional/glusterd/test_readonly_option_on_volume.py
```diff
@@ -0,0 +1,62 @@
...
+        3. set 'read-only on' on the volume
+        4. perform some I/O's on mount point
+        5. set 'read-only off' on the volume
+        6. perform some I/O's on mount point
+        """
+        error_msg = "Read-only file system"
```
nik-redhat, line 40: Is this msg applicable for all OS? Like any specific reason for checking this error msg?

common/ops/support_ops/io_ops.py
```diff
@@ -27,7 +27,8 @@ def create_file(self, path: str, filename: str, node: str):
         cmd = f"touch {path}/{filename}"
         self.execute_abstract_op_node(cmd, node)
 
-    def create_dir(self, path: str, dirname: str, node: str):
+    def create_dir(self, path: str, dirname: str, node: str,
+                   excep: bool = True):
```
nik-redhat, line 31: Just thinking by the wording logic, shouldn't this be `false` by default., considering this will have to be changed in the asbtrac_op_node() as well.

tests/functional/glusterd/test_readonly_option_on_volume.py
srijan-sivakumar: Now, in RHEL and Fedora based systems, this is a standard wording. Coming to debian systems or even suse and other distros, I'm not sure of it. One way of handling this is then simply checking if the IO failed but that would be in-adequate and this is how it is done in glusto. Hence seeing that the commonly used distro is either fedora, centos or RHEL systems, I took this as the form for validation.

common/ops/support_ops/io_ops.py
srijan-sivakumar: So excep stands for do we take exception handling in the abstract ops. By default it is true. Exception handling has to be taken care of inside the abstract ops to judge whether a command fails or passes.

nik-redhat: Are... I meant by the docstring which is added in the `execute_Abstract_ops_node` based on that the working seems to be opposite. No problem with the logic as such.

nik-redhat: Anyway, let's not break our heads on this :p

nik-redhat (approved):

Issue #373, opened 2021-05-18
[Tools] File operations script to be sent across to clients


srijan-sivakumar:
The file operations script present in glusto-tests is useful for performing file IO operations and instead of re-writing the same, we can reuse the script. The script in question is : [file_dir_ops.py](https://github.com/gluster/glusto-tests/blob/master/glustolibs-io/shared_files/scripts/file_dir_ops.py).

This script can be copied to the /tmp directory of the clients and used as and when required.

PR #422, opened 2021-05-24
[Core][Common][Test][Tools] File_dir_script

srijan-sivakumar:
Following changes have been done
    1. file_dir_ops migrated from glusto to be used for redant.
    2. Environ handles the transfer of the file_dir_ops from
    the tools.
    3. Mixin constructor parameters changed to reflect
    change in how the client and server list is handled.
    4. Correspondingly, the Rexe constructor is also
    modified to obtain server and client list separately.

Fixes: #373 


nik-redhat:
Having a function like `start_io` in io_ops might be helpful, which invokes that script?

srijan-sivakumar:
Yes...but I'm taking that as a separate patch. We will create abstractions depending on the functionality invoked from that script.

nik-redhat (approved):
lgtm

Issue #418, opened 2021-05-24
Clearing out volume in teardown for disruptive TCs


nishith-vihar:
If for suppose the disruptive test case contains volume creation as a part of the test case logic and the test case fails midway the volume created as a part of the test case has to be deleted in the teardown in dParentTest making use of the 'volds' data structure.

PR #426, opened 2021-05-24
Cleanup of volumes creates in test case in disruptive tc

nishith-vihar:
In case of disruptive TC if the test_case fails midway then
the volumes created in the tc are deleted

Fixes: #418


srijan-sivakumar (approved):
LGTM

Issue #408, opened 2021-05-20
Teardown enhancement, volds and cleanup


srijan-sivakumar:
All these three factors, teardown, volds, cleands and cleanup come together in the parent test cases. Their handling has to be improved to solve certain corner cases which can arise when test cases fail.

srijan-sivakumar:
- [ ]  Adding couple of functions which would be required in coming times, `sanitize_volume(volname)` and `cleanup_volume(volname)` wherein sanitizing would consist of

1. Cleaning up of mount if mounted
2. Starting the volume if not started
3. Mounting the volume if not mounted
4. creating the volume if it had been deleted for some reason by the test case.

- [ ] Following this addition one more addition would be required, cleaning up of the bricks in case of volume deletion.
- [ ] Handling test case specific volume list to handle them in teardowns if the test case failed.

PR #428, opened 2021-05-24
[Core][Common][Test] Sanitize and cleanup

srijan-sivakumar:
Sanitize and cleanup functionalities added.
1. The volume sanitize is used for prepping a volume for the next TC or even to be used within a test case.
2. The volume cleanup is to be used for complete cleanup of volume and its mount.

Also added is the brick dir deletion which is handled at
1. nd_parent_test
2. d_parent_test
3. vol_destroy_test

Fixes: #408
    
### DIFF
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -348,3 +348,14 @@ class BrickOps:
             server_iter += 1
 
         return (brick_dict, brick_cmd)
+
+    def cleanup_brick_dirs(self):
+        """
+        This function requests for the cleands and clears the brick dirs which
+        have been populated in it.
+        """
+        cleands_val = list(self.es.get_cleands_data().items())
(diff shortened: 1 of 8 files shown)


srijan-sivakumar:
@nik-redhat @nishith-vihar @neesingh-rh @aujjwal-redhat this is ready for review now.

aujjwal-redhat (approved):
LGTM

nishith-vihar:
There are 2 things. First one being in line no: 52 of nd_parent_test the self.test_name variable represents test case name+vol type. Is it possible to change it in a way that it holds only test case name?
Second one is that d_parent_test does not hold a test_name. I think its better to add a self.test_name variable so that it is helpful in naming the volume names being created inside the test case

srijan-sivakumar:
@nishith-vihar will handle the nomenclature of test and volumes in a separate patch.

nishith-vihar (approved):
LGTM

Issue #419, opened 2021-05-24
[TEST] Migration of test_change_reserve_limit.py


nishith-vihar:
Migration of [test_case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_change_reserve_limit.py)

PR #433, opened 2021-05-25
Migration of test_change_reserve_limit.py

nishith-vihar:
Migration of [test_case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_change_reserve_limit.py)

Updates: #292
Fixes: #419 


nishith-vihar:
The current implementation of remove_brick does not take in as parameter the number of bricks to remove from the test case.  The remove_brick parts have been commented in the test case and has to be taken up once the implementation of the remove_brick is done in brick_libs.

srijan-sivakumar:
@nishith-vihar I changed the code but am not able to push the code to this branch...probably I will open a new PR...that'll also contain your commits...

Issue #445, opened 2021-05-25
[TEST] Migration of test_brick_port_after_stop_glusterd_modify_volume.py


nishith-vihar:
Migration of [test_case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py) from glusto to redant

PR #448, opened 2021-05-25
Migration of test_brick_port_after_stop_glusterd_modify_volume.py

nishith-vihar:
Migration of [test_case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py)

Updates: #292
Fixes: #445

### DIFF
--- a/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
+++ b/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
@@ -28,8 +28,14 @@ from tests.d_parent_test import DParentTest
 class TestCase(DParentTest):
 
     def terminate(self):
+        """
+        In case the test fails midway and one of the nodes has
+        glusterd stopped then the glusterd is started on that node
+        and then the terminate function in the DParentTest is called
+        """
         if self.glusterd_stopped:
             self.redant.start_glusterd(self.server_list[1])
+            self.redant.wait_for_glusterd_to_start(self.server_list[1])
         super().terminate()
 
     def run_test(self, redant):


tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,105 @@
...
+
+        vol_bricks = vol_status[self.vol_name]['node']
+        if not isinstance(vol_bricks, list):
+            vol_bricks = [vol_bricks]
+        totport = 0
+        for vol_brick in vol_bricks:
+            if vol_brick["port"] == 'N/A':
+                continue
+            if int(vol_brick["port"]) > 0:
+                totport += 1
```
nik-redhat, line 57: Are you sure, this is working after changing dict to list? Can you attach the successful run output as well?

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,105 @@
...
+from time import sleep
+from tests.d_parent_test import DParentTest
+
+
+# disruptive;dist
+class TestCase(DParentTest):
+
+    def terminate(self):
+        if not self.redant.is_glusterd_running(self.server_list[1]):
+            self.redant.stop_glusterd(self.server_list[0])
```
srijan-sivakumar, line 27: @nishith-vihar  just curious..why are we stopping glusterd if it is not working..

srijan-sivakumar (changes requested):
I think we need to start glusterd here if it is not running and also, keep a check condition in such a manner that we will wait till it is started, something like wait till glusterd is stopped.

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,105 @@
...
+from time import sleep
+from tests.d_parent_test import DParentTest
+
+
+# disruptive;dist
+class TestCase(DParentTest):
+
+    def terminate(self):
+        if not self.redant.is_glusterd_running(self.server_list[1]):
+            self.redant.start_glusterd(self.server_list[0])
```
srijan-sivakumar, line 27: let's keep a call here for `wait_till_glusterd_is_up` ( you can check it in gluster_ops.py ). This will prevent the terminate operations from doing something before the said node's gluster daemon comes up.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,105 @@
...
+
+        ret = redant.wait_for_peers_to_connect(self.server_list[0],
+                                               self.server_list[1])
+        if not ret:
+            raise Exception("glusterd is not connected "
+                            f"{self.server_list[0]} with peer "
+                            f"{self.server_list[1]}")
+
+        # Waiting for 10 sec so that the brick will get port
+        sleep(10)
```
nik-redhat, line 85: Does it really take so long, to assign a brick port :thinking:

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,105 @@
...
+
+        vol_bricks = vol_status[self.vol_name]['node']
+        if not isinstance(vol_bricks, list):
+            vol_bricks = [vol_bricks]
+        totport = 0
+        for vol_brick in vol_bricks:
+            if vol_brick["port"] == 'N/A':
+                continue
+            if int(vol_brick["port"]) > 0:
+                totport += 1
```
nik-redhat, line 100: Same as above

nik-redhat (changes requested):

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nik-redhat: Also, I don't get the usage? You are checking glusterd on node[1] is stopped and starting it on node[0]?
Secondly, we don't need a check like this. Instead, use a class variable like `self.glusterd_stopped = True` and then update it after a start and check this variable for starting glusterd (if needed) in terminate.

nik-redhat (changes requested):

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nishith-vihar: @srijan-sivakumar Ack

nishith-vihar: @nik-redhat Yes its like the teardown in glusto

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+from time import sleep
+from tests.d_parent_test import DParentTest
+
+
+# disruptive;dist
+class TestCase(DParentTest):
+
+    def run_test(self, redant):
+        # pylint: disable=too-many-statements, too-many-branches
```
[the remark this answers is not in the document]
nishith-vihar: Ack

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -0,0 +1,101 @@
...
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License along
+  with this program; if not, write to the Free Software Foundation, Inc.,
+  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
```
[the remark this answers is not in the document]
nishith-vihar: Ack

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nishith-vihar: I saw the structure of volume status dict and wrote the logic. I will attach the screenshot for reference.

nik-redhat:
Why are we excluding this set of functionality?
https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py#L85-L111

Though, one update here would be to not run the test on any set of volume (which I think makes it 'Gnereic'? @srijan-sivakumar ), since we are just creating a volume within the TC and using it.

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nishith-vihar: Test case is failing with 5 seconds. With 10 seconds its passing.

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nishith-vihar: Ack

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
nik-redhat: Did you try multiple times?

nishith-vihar:
> Why are we excluding this set of functionality?
> https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py#L85-L111
> 
> Though, one update here would be to not run the test on any set of volume (which I think makes it 'Gnereic'? @srijan-sivakumar ), since we are just creating a volume within the TC and using it.

This is creation and starting of volume which is already done as a part of the cluster creation

tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
srijan-sivakumar: > @srijan-sivakumar is this `terminate()` like a teardown function?

Somewhat similar. So basically operations like cleanup ( basic level of cleanup which is generic to all test cases ) is part of the terminate inside the parent_test. But coming to scenarios like the one we have here, wherein we are stopping glusterd, we need to start glusterd before the test case ends. Now, we could've added this check inside the terminate in parent test itself but this isn't very generic. Hence the idea to add it in the terminate of the test case.

srijan-sivakumar: > Also, I don't get the usage? You are checking glusterd on node[1] is stopped and starting it on node[0]?
> Secondly, we don't need a check like this. Instead, use a class variable like `self.glusterd_stopped = True` and then update it after a start and check this variable for starting glusterd (if needed) in terminate.

@nishith-vihar , now that I see, we are actually doing the check in a node and starting it in some other node.

nishith-vihar: @nik-redhat First thing, my mistake, it should have been node[1] in start_glusterd. Second thing, i think the point about `self.glusterd_stopped=True` is a valid point and seems fine. I can implement that way only

nishith-vihar:
> > > Why are we excluding this set of functionality?
> > > https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py#L85-L111
> > > Though, one update here would be to not run the test on any set of volume (which I think makes it 'Gnereic'? @srijan-sivakumar ), since we are just creating a volume within the TC and using it.
> > 
> > 
> > This is creation and starting of volume which is already done as a part of the cluster creation
> 
> TBH.....it works for this TC, but that might not be the scenario for all of them. So, do check once before ignoring it, that how the TC is working.

Yeah nikhil. After i check that a part of the tc is already being done in cluster creation, then only i am removing that part.

nishith-vihar:
Screenshot from 2021-05-26 12-59-51

common/ops/gluster_ops/gluster_ops.py
```diff
@@ -228,6 +227,27 @@ def wait_for_glusterd_to_start(self, node: str, timeout: int = 80) -> bool:
             count += 1
         return False
 
+    def wait_for_glusterd_to_stop(self, node: str, timeout: int = 80) -> bool:
```
nik-redhat, line 230: Add this function in the docs as well.

nishith-vihar: You mean in the Steps for the tc?

nik-redhat: No, I mean the docs for the `ops` functions which you guys added a few days back.

nik-redhat (approved):

Issue #410, opened 2021-05-21
[TEST] Migrate test_reserved_port_range_for_gluster


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_reserved_port_range_for_gluster.py)

PR #458, opened 2021-05-26
Migrated test_reserved_port_range_for_gluster

aujjwal-redhat:
Migrated the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_reserved_port_range_for_gluster.py)

Fixes: #410

Updates: #292

### DIFF
--- a/tests/functional/glusterd/test_reserved_port_range_for_gluster.py
+++ b/tests/functional/glusterd/test_reserved_port_range_for_gluster.py
@@ -1,22 +1,22 @@
-#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License along
-#  with this program; if not, write to the Free Software Foundation, Inc.,
-#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-
-""" Description:
-      Setting reserved port range for gluster
 """
+Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+Description:
+  Setting reserved port range for gluster
 
 from random import choice
 from glusto.core import Glusto as g
@@ -28,32 +28,31 @@ from glustolibs.gluster.volume_libs import cleanup_volume
 from glustolibs.gluster.lib_utils import get_servers_bricks_dict
 from glustolibs.gluster.gluster_init import restart_glusterd
 from glustolibs.gluster.peer_ops import wait_for_peers_to_connect
+"""
+
+# disruptive;
+
+from tests.d_parent_test import DParentTest
+from time import sleep
 
 
-class TestReservedPortRangeForGluster(GlusterBaseClass):
-    def tearDown(self):
+class TestCase(DParentTest):
+
+    def terminate(self):
+        """
+        In case the test fails in midway
+        then the port range needs to be brought
+        back to the previous state.
+        """
         # Reset port range if some test fails
         if self.port_range_changed:
-            cmd = "sed -i 's/49200/60999/' /etc/glusterfs/glusterd.vol"
-            ret, _, _ = g.run(self.mnode, cmd)
-            self.assertEqual(ret, 0, "Failed to set the max-port back to"
-                             " 60999 in glusterd.vol file")
+            cmd = ("sed -i 's/49200/60999/' "
+                   "/usr/local/etc/glusterfs/glusterd.vol")
+            self.redant.execute_abstract_op_node(cmd, self.server_list[0])
 
-        # clean up all volumes
-        vol_list = get_volume_list(self.mnode)
-        if vol_list is None:
-            raise ExecutionError("Failed to get the volume list")
+        super().terminate()
 
-        for volume in vol_list:
-            ret = cleanup_volume(self.mnode, volume)
-            if not ret:
-                raise ExecutionError("Unable to delete volume %s" % volume)
-            g.log.info("Volume deleted successfully : %s", volume)
-
-        # Calling baseclass tearDown method
-        self.get_super_method(self, 'tearDown')()
-
-    def test_reserved_port_range_for_gluster(self):
+    def run_test(self, redant):
         """
         Test Case:
         1) Set the max-port option in glusterd.vol file to 49200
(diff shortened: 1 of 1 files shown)


aujjwal-redhat:
@nik-redhat @srijan-sivakumar In this test case, 50 volumes are created but only 24 are able to start. It fails for 25th volume with the message `Commit failed on localhost. Please check log file for details.`

It is running perfectly for 25 volume scenario but not fully for 50 volumes.

nik-redhat:
You can use this script:
```
for i in {1..75}
do
    gluster vol create vol$i <brick-list>
    sleep 1
    gluster vol start vol$i
done
```

aujjwal-redhat:
> > @nik-redhat @srijan-sivakumar In this test case, 50 volumes are created but only 24 are able to start. It fails for 25th volume with the message `Commit failed on localhost. Please check log file for details.`
> > It is running perfectly for 25 volume scenario but not fully for 50 volumes.
> 
> Try creating 50 volumes in that setup, using a bash script. Without the framework. I think it's a memory issue.

50 volumes are getting created but start is failing.

nik-redhat:
Using, bash as well?

aujjwal-redhat:
> You can use this script:
> 
> ```
> for i in {1..75}
> do
>     gluster vol create vol$i <brick-list>
>     sleep 1
>     gluster vol start vol$i
> done
> ```

yeah will have to try something like this

aujjwal-redhat:
> Using, bash as well?

Ahh.. no just from the framework.

50 vol gets created, start goes up till 24 and then commit fails and deletes the 50 vols

nik-redhat:
Try from bash, I think you will get the same result or at max 1-5 volumes extra.

srijan-sivakumar:
I think it is the memory and the env issue. Gluster is not meant for creating 50 volumes in 3 VMs that too very light weight VMs.

srijan-sivakumar:
Such a test case cannot be run in the VMs we are using. ( Also a customer should be sane enough to not to try such things in a small resource constrained VM ) Technically it is not even a recommended operation.

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+    def terminate(self):
+        """
+        In case the test fails in midway
+        then the port range needs to be brought
+        back to the previous state.
+        """
+        # Reset port range if some test fails
+        if self.port_range_changed:
+            cmd = ("sed -i 's/49200/60999/' "
+                   "/usr/local/etc/glusterfs/glusterd.vol")
```
srijan-sivakumar, line 39: the path has to be changed to `/etc/glusterfs/glusterd.vol`

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+        6) Confirm the error message, due to which volume failed to start
+        7) Set the max-port option in glusterd.vol file back to default value
+        8) Restart glusterd on the same node
+        9) Starting the 50th volume should succeed now
+        """
+        try:
+
+            # Set max port number as 49200 in glusterd.vol file
+            cmd = ("sed -i 's/60999/49200/' "
+                   "/usr/local/etc/glusterfs/glusterd.vol")
```
srijan-sivakumar, line 61: path has to be changes to `/etc/glusterfs/glusterd.vol`

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+            ret = redant.execute_abstract_op_node(cmd, self.server_list[0])
+
+            out = ret['msg'][0].rstrip('\n')
+            if int(out) == 0:
+                raise Exception("Volume start didn't fail with expected"
+                                " error message")
+
+            # Set max port number back to default value in glusterd.vol file
+            cmd = ("sed -i 's/49200/60999/' "
+                   "/usr/local/etc/glusterfs/glusterd.vol")
```
srijan-sivakumar, line 117: same path change.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+
+            # Set max port number as 49200 in glusterd.vol file
+            cmd = ("sed -i 's/60999/49200/' "
+                   "/etc/glusterfs/glusterd.vol")
+            redant.execute_abstract_op_node(cmd, self.server_list[0])
+
+            self.port_range_changed = True
+            # Restart glusterd
+            redant.restart_glusterd(self.server_list[0])
+            sleep(2)
```
srijan-sivakumar, line 67: Also.instead of sleep let's use wait_for_glusterd_to_start function...

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+            # Set max port number back to default value in glusterd.vol file
+            cmd = ("sed -i 's/49200/60999/' "
+                   "/etc/glusterfs/glusterd.vol")
+            redant.execute_abstract_op_node(cmd, self.server_list[0])
+
+            self.port_range_changed = False
+
+            # Restart glusterd on the same node
+            redant.restart_glusterd(self.server_list[0])
+
```
srijan-sivakumar, line 124: call the wait_for_glusterd_to_start here also.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,133 @@
...
+        2) Restart glusterd on one of the node
+        3) Create 50 volumes in a loop
+        4) Try to start the 50 volumes in a loop
+        5) Confirm that the 50th volume failed to start
+        6) Confirm the error message, due to which volume failed to start
+        7) Set the max-port option in glusterd.vol file back to default value
+        8) Restart glusterd on the same node
+        9) Starting the 50th volume should succeed now
+        """
+        try:
```
srijan-sivakumar, line 57: Also..we don't need try-except over the whole test...we can remove that.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Yeah, that was while testing. Will remove it.

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,131 @@
...
+        7) Set the max-port option in glusterd.vol file back to default value
+        8) Restart glusterd on the same node
+        9) Starting the 50th volume should succeed now
+        """
+        # Set max port number as 49200 in glusterd.vol file
+        cmd = ("sed -i 's/60999/49200/' "
+               "/etc/glusterfs/glusterd.vol")
+        redant.execute_abstract_op_node(cmd, self.server_list[0])
+
+        self.port_range_changed = True
```
srijan-sivakumar, line 61: nit: Do put a variable definition above the command execution on line 56. `self.port_range_changed = False`. The reason ? => Well if the execute_abstract_op_node command failed, we will go to terminate and as `self.port_range_changed` has not yet been declared, it will throw an exception. So let's just add the initial value as false and put it on line 56, before the command execution.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
aujjwal-redhat: Ack

srijan-sivakumar (approved):
LGTM

tests/functional/glusterd/test_reserved_port_range_for_gluster.py
```diff
@@ -0,0 +1,136 @@
...
+        if not redant.wait_for_peers_to_connect(self.server_list,
+                                                self.server_list[1]):
+            raise Exception("Peers node in connected mode.")
+
+        # Create 50 volumes in a loop
+        for i in range(1, 51):
+            volname = f"{self.vol_name}-volume-{i}"
+            redant.volume_create(volname, self.server_list[0],
+                                 self.vol_type_inf[self.conv_dict['dist']],
+                                 self.server_list, self.brick_roots, True)
```
nik-redhat, line 82: Why is volume created with `force`?

srijan-sivakumar: Actually the default volume creation itself is force if you remember in the CI env.. :p

srijan-sivakumar: Also..I think it is volume start which is to fail..in this test case..so would volume creation with force affect that ? ( Technically the volume 50 isn't created as ports have been exhausted )

nik-redhat: Volume create would succeed with the port restriction, but the start would fail.
Yes, I remember the CI limitation, but let's not have the same being done in the TC, the test doesn't require the volume to be created with force. So, that when we fix the CI and do not do force create by default the TC actually works as expected.

srijan-sivakumar: Cool..then we can remove the force in this case...

Issue #463, opened 2021-05-26
Check for glusterd running to be removed in the test_brick_port_after_stop_glusterd_modify_volume.py


nishith-vihar:
glusterd running check has to be removed in the test case and all the functionalities except calling the super function has to be placed in the try except block.

PR #464, opened 2021-05-26
terminate funtionalities are added in try except

nishith-vihar:
terminate funtionalities are added in try except and self.gusterd_stopped class variable is removed

Fixes: #463

### DIFF
--- a/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
+++ b/tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
@@ -20,6 +20,7 @@
   making changes on the volume and starting the glusterd
   again.
 """
+import traceback
 from time import sleep
 from tests.d_parent_test import DParentTest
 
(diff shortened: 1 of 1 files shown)


tests/functional/glusterd/test_brick_port_after_stop_glusterd_modify_volume.py
```diff
@@ -33,9 +34,13 @@ def terminate(self):
         glusterd stopped then the glusterd is started on that node
         and then the terminate function in the DParentTest is called
         """
-        if self.glusterd_stopped:
+        try:
             self.redant.start_glusterd(self.server_list[1])
             self.redant.wait_for_glusterd_to_start(self.server_list[1])
+        except Exception as error:
+            tb = traceback.format_exc()
```
nik-redhat, line 41: I don't think we need a try block here, itself.

srijan-sivakumar: No no..let's keep it here. If due to some reason start_glusterd failed, we need to handle it.

srijan-sivakumar (approved):

Issue #465, opened 2021-05-26
[TEST] Migration of test_volume_status_with_absent_bricks


neesingh-rh:
Migrate test case:[test_volume_status_with_absent_bricks](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_volume_status_with_absent_bricks.py)

aujjwal-redhat:
This test case is merged already. Apparently, instead of updates, fixes should have been used. Hence, closing this issue manually.

PR #466, opened 2021-05-26
Migration of test_volume_status_with_absent_bricks

neesingh-rh:
Migrated [test case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_volume_status_with_absent_bricks.py)

Updates: #465 

Signed-off-by: Neeraj Pratap Singh <neesingh@redhat.com>

### DIFF
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -173,7 +173,7 @@ class VolumeOps(AbstractOps):
                 will get executed with force option. If it is set to False,
                 then start volume will get executed without force option
             excep (bool): To bypass or not to bypass the exception handling.
-            
+
         Returns:
             ret: A dictionary consisting
                 - Flag : Flag to check if connection failed
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -172,7 +172,6 @@ class VolumeOps(AbstractOps):
             force (bool): If this option is set to True, then start volume
                 will get executed with force option. If it is set to False,
                 then start volume will get executed without force option
-            excep (bool): To bypass or not to bypass the exception handling.
 
         Returns:
             ret: A dictionary consisting
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -172,6 +172,8 @@ class VolumeOps(AbstractOps):
             force (bool): If this option is set to True, then start volume
                 will get executed with force option. If it is set to False,
                 then start volume will get executed without force option
+            excep (bool): To bypass or not to bypass the exception handling.
+            
         Returns:
             ret: A dictionary consisting
                 - Flag : Flag to check if connection failed
--- a/tests/functional/glusterd/test_volume_status_with_absent_bricks.py
+++ b/tests/functional/glusterd/test_volume_status_with_absent_bricks.py
@@ -59,7 +59,7 @@ class TestCase(DParentTest):
         # Checking volume status
         ret = redant.get_volume_status(self.vol_name, self.server_list[0],
                                        excep=False)
-        if ret['error_code'] != 0 and ret['msg']['opErrstr'] != \
-        f'Volume {self.vol_name} is not started':
+        if ret['error_code'] != 0 and ret['msg']['opErrstr'] !=\
+           f'Volume {self.vol_name} is not started':
             raise Exception("Incorrect error message for gluster vol "
                             "status")
--- a/tests/functional/glusterd/test_volume_status_with_absent_bricks.py
+++ b/tests/functional/glusterd/test_volume_status_with_absent_bricks.py
@@ -1,0 +1,118 @@
+#  Copyright (C) 2018  Red Hat, Inc. <http://www.redhat.com>
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+""" Description:
+      Volume start when one of the brick is absent
+"""
+
+import random
+from glusto.core import Glusto as g
+from glustolibs.gluster.exceptions import ExecutionError
+from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
+from glustolibs.gluster.volume_ops import (volume_create, volume_start,
+                                           volume_status)
+from glustolibs.gluster.lib_utils import form_bricks_list
+
+
+@runs_on([['distributed', 'replicated', 'distributed-replicated'],
+          ['glusterfs']])
+class TestVolumeStatusWithAbsentBricks(GlusterBaseClass):
+
+    def tearDown(self):
+        """
+        tearDown for every test
+        """
+        # stopping the volume and Cleaning up the volume
+        ret = self.cleanup_volume()
+        if not ret:
+            raise ExecutionError("Failed Cleanup the Volume %s"
+                                 % self.volname)
+        # Calling GlusterBaseClass tearDown
+        GlusterBaseClass.tearDown.im_func(self)
+
+    def test_volume_absent_bricks(self):
+        '''
+        -> Create Volume
+        -> Remove any one Brick directory
+        -> Start Volume
+        -> Check the gluster volume status
+        '''
+        num_of_bricks = 0
+        replica = True
+
+        if self.volume_type == 'distributed':
+            num_of_bricks = 3
+            replica = False
+
+        elif self.volume_type == 'replicated':
+            num_of_bricks = 3
+
+        elif self.volume_type == 'distributed-replicated':
+            num_of_bricks = 6
+
+        # Forming brick list
+        brick_list = form_bricks_list(self.mnode, self.volname, num_of_bricks,
+                                      self.servers, self.all_servers_info)
+        if replica:
+            # Creating Volume
+            ret, _, _ = volume_create(self.mnode, self.volname, brick_list,
+                                      replica_count=3)
+            self.assertEqual(ret, 0, "Volume creation failed for %s"
+                             % self.volname)
+            g.log.info("volume created successfully %s", self.volname)
+        else:
+            # Creating Volume
+            ret, _, _ = volume_create(self.mnode, self.volname, brick_list)
+            self.assertEqual(ret, 0, "Volume creation failed for %s"
+                             % self.volname)
+            g.log.info("volume created successfully %s", self.volname)
+
+        # Command for removing brick directory
+        random_brick = random.choice(brick_list)
+        node, brick_path = random_brick.split(r':')
+        cmd = 'rm -rf ' + brick_path
+
+        # Removing brick directory of one node
+        ret, _, _ = g.run(node, cmd)
+        self.assertEqual(ret, 0, "Failed to remove brick dir")
+        g.log.info("Brick directory removed successfully")
+
+        # Starting volume
+        ret, _, err = volume_start(self.mnode, self.volname)
+        self.assertNotEqual(ret, 0, "Unexpected: Volume started successfully "
+                                    "even though brick directory is removed "
+                                    "for %s" % self.volname)
+        g.log.info("Expected: Failed to start volume %s", self.volname)
+
+        # Checking volume start failed message
+        msg = "Failed to find brick directory"
+        self.assertIn(msg, err, "Expected message is %s but volume start "
+                                "command failed with this "
+                                "message %s" % (msg, err))
+        g.log.info("Volume start failed with correct error message %s", err)
+
+        # Checking Volume status
+        ret, _, err = volume_status(self.mnode, self.volname)
+        self.assertNotEqual(ret, 0, "Success in getting volume status, volume "
+                                    "status should fail when volume is in "
+                                    "not started state ")
+        g.log.info("Failed to get volume status which is expected")
+
+        # Checking volume status message
+        msg = ' '.join(['Volume', self.volname, 'is not started'])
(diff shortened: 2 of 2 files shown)


tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Removing any one brick directory
+        brick_list = list(brick_list.items())
+        random_brick = random.choice(brick_list)
+        random_brick_path = random.choice(random_brick[1])
+        cmd = f'rm -rf {random_brick_path}'
+        redant.execute_abstract_op_node(cmd, random_brick[0])
+        redant.logger.info("Brick directory removed successfully")
+
+        redant.volume_stop(self.vol_name, self.server_list[0])
```
nik-redhat, line 50: The above operations need to be done, with volume stopped. Check the setup part of the glusto TC.
So, we won't be needing a volume_stop here.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        # Starting volume
+        err_msg = 'Failed to find brick directory'
+        cmd = f"gluster volume start {self.vol_name}"
```
nik-redhat, line 53: Use the volume_start function.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        ret = redant.execute_command(cmd, self.server_list[0])
```
nik-redhat, line 54: USe abstract function. The `excep` param needs to be added in the `volume_start` function as well. For example, check implementation of `volume_unmount` function.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        if err_msg not in ret['error_msg']:
+            raise Exception("Unexpected!!")
```
nik-redhat, line 56: What is `Unexpected!!`?

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        # Checking volume status
+        cmd = f'gluster vol status {self.vol_name}'
+        ret = redant.execute_command(cmd, self.server_list[0])
+        if ret['error_msg'] != f'Volume {self.vol_name} is not started\n':
+            raise Exception("Volume status erraneous")
+
+        try:
+            redant.get_volume_status(self.vol_name, self.server_list[0])
+        except:
+            redant.logger.info("Volume hasn't started hence no status for it.")
```
nik-redhat, line 72: Why are you doing the same operation twice?

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+            redant.volume_start(self.vol_name, self.server_list[0])
+        except:
+            redant.logger.info(f"Expected: Failed to start volume"
+                               f"{self.vol_name}")
+
+        # Checking volume status
+        cmd = f'gluster vol status {self.vol_name}'
+        ret = redant.execute_command(cmd, self.server_list[0])
+        if ret['error_msg'] != f'Volume {self.vol_name} is not started\n':
+            raise Exception("Volume status erraneous")
```
nik-redhat, line 67: Check the error message from glusto TC, use the same one.

nik-redhat (changes requested):

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+    def run_test(self, redant):
+        """
+        Test Case:
+        1) Create Volume
+        2) Remove any one Brick directory
+        3) Start Volume and compare the failure message
+        4) Check the gluster volume status and compare the status message
+        """
+        # Fetching the brick list
+        redant.logger.info("Fetch all the bricks of the volume")
```
srijan-sivakumar, line 37: We can remove this log. I don't think this is needed.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        brick_list = redant.es.get_brickdata(self.vol_name)
```
srijan-sivakumar, line 38: The data structure will be containing the bricks to be provided, in glusto the approach used to obtain the bricks is to query the servers which isn't being done here. Hence we can remove the check for it and also raising the exception.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        if brick_list is None:
+            raise Exception("Failed to get the brick list")
+
+        # Removing any one brick directory
+        brick_list = list(brick_list.items())
+        random_brick = random.choice(brick_list)
+        random_brick_path = random.choice(random_brick[1])
+        cmd = f'rm -rf {random_brick_path}'
+        redant.execute_abstract_op_node(cmd, random_brick[0])
+        redant.logger.info("Brick directory removed successfully")
```
srijan-sivakumar, line 48: We don't need the success log.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
srijan-sivakumar: Yes, we need to stop the volume at the very beginning itself.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,72 @@
...
+        if err_msg not in ret['error_msg']:
+            raise Exception("Unexpected!!")
+        try:
+            redant.volume_start(self.vol_name, self.server_list[0])
+        except:
+            redant.logger.info(f"Expected: Failed to start volume"
+                               f"{self.vol_name}")
+
+        # Checking volume status
+        cmd = f'gluster vol status {self.vol_name}'
```
srijan-sivakumar, line 64: I think we have a volume_status function in Redant. Please use that. The idea is to use the ops functions whenever they're available and not to recreate their functionality again and again in test cases.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: That is the case when the error message is other than "failed to find brick directory" which is unexpected.I will complete it

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: first one for raising exception err_mssg is other than the expected one..second one is just for logging(I guess its not required).My bad.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: ACK

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: ack

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: Same here,will change

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: Ack

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: Ack

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,60 @@
...
+    def run_test(self, redant):
+        """
+        Test Case:
+        1) Create Volume
+        2) Remove any one Brick directory
+        3) Start Volume and compare the failure message
+        4) Check the gluster volume status and compare the status message
+        """
+        # Fetching the brick list
+        redant.volume_stop(self.vol_name, self.server_list[0])
```
nik-redhat, line 37: Move this above the `Fetching brick list` comment.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,60 @@
...
+        brick_list = redant.es.get_brickdata(self.vol_name)
```
nik-redhat, line 38: Use the function `get_all_bricks`. Check the [brick_ops_doc](https://github.com/srijan-sivakumar/redant/blob/main/docs/BP/Ops/brick_ops.md) for refeerence.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # Removing any one brick directory
+        brick_list = list(brick_list.items())
+        random_brick = random.choice(brick_list)
+        random_brick_path = random.choice(random_brick[1])
```
nik-redhat, line 43: It should be split instead, please check the working after updating the above code to use get_all_bricks

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,60 @@
...
+        cmd = f'rm -rf {random_brick_path}'
+        redant.execute_abstract_op_node(cmd, random_brick[0])
+
+        # Starting volume
+        err_msg = 'Failed to find brick directory'
+        ret = redant.volume_start(self.vol_name, self.server_list[0],
+                                  excep=False)
+        if err_msg not in ret['msg']['opErrstr']:
```
nik-redhat, line 51: Before this check, do a check for the `error_code` == 0

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,60 @@
...
+            raise Exception("Unexpected:Volume started successfully"
+                            " even though brick is deleted.")
+
+        # Checking volume status
+        ret = redant.get_volume_status(self.vol_name, self.server_list[0],
+                                       excep=False)
+        if ret['msg']['opErrstr'] != f'Volume {self.vol_name} is not started':
```
nik-redhat, line 58: Add a check before this check for `error_code`.

nik-redhat, line 58: Also, it might require `rstrip('\n')` have you confirmed the working of the TC?

nik-redhat (changes requested):
Please add test run result after making the changes, and only then request for re-review.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
neesingh-rh: ACK

neesingh-rh: Its not required.

neesingh-rh:
image

common/ops/gluster_ops/volume_ops.py
```diff
@@ -166,6 +166,8 @@ def volume_start(self, volname: str, node: str = None,
             force (bool): If this option is set to True, then start volume
                 will get executed with force option. If it is set to False,
                 then start volume will get executed without force option
+            excep (bool): To bypass or not to bypass the exception handling.
```
nik-redhat, line 169: not needed, it has been added above already.

srijan-sivakumar:
@aujjwal-redhat just test the case once so as to re-assure with latest changes.

aujjwal-redhat:
> @aujjwal-redhat just test the case once so as to re-assure with latest changes.

yeah running already. waiting for the results.

aujjwal-redhat:
```js
Summary:
Table:
 test_volume_status_with_absent_bricks
 
|     dist    |     PASS    | 23.794852018356323 |
|     rep     |     PASS    | 23.931774377822876 |
|   dist-rep  |     PASS    | 29.602852821350098 |
|     disp    |     PASS    | 29.248443603515625 |
|  dist-disp  |     PASS    | 37.676634788513184 |
|     arb     |     PASS    | 24.161861419677734 |



Framework runtime : 207.5736448764801
```

srijan-sivakumar (approved):

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,65 @@
...
+        random_brick = random.choice(brick_list)
+        node, brick_path = random_brick.split(':')
+        cmd = 'rm -rf ' + brick_path
+        redant.execute_abstract_op_node(cmd, node)
+
+        # Starting volume
+        err_msg = 'Failed to find brick directory'
+        ret = redant.volume_start(self.vol_name, self.server_list[0],
+                                  excep=False)
+        if ret['error_code'] != 0 and err_msg not in ret['msg']['opErrstr']:
```
nik-redhat, line 55: It would have been better to check the `ret['msg']['opRet']` value as well, `error_code` doesn't always give the correct value.

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
```diff
@@ -0,0 +1,65 @@
...
+            raise Exception("Unexpected:Volume started successfully"
+                            " even though brick is deleted.")
+
+        # Checking volume status
+        ret = redant.get_volume_status(self.vol_name, self.server_list[0],
+                                       excep=False)
+        if ret['error_code'] != 0 and ret['msg']['opErrstr'] !=\
```
nik-redhat, line 62: Same as above

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
aujjwal-redhat: Ok I will submit a patch. : )
Thanks @nik-redhat for the review

tests/functional/glusterd/test_volume_status_with_absent_bricks.py
aujjwal-redhat: Ack

Issue #467, opened 2021-05-27
Mountpoint Specific ops changes and cleanup


srijan-sivakumar:
Also addressing in this patch is the following function
1. Replacing `ls` with `stat` inside the `path_exists` function.
2.  if_mountpoint_exists -> This is required to see if the mountpoint exists at the given path. This function will be encapsulating the path_exists function and based on the response of the path_exists it will determine whether the mountpoint doesn't even exists or if it exists and the volume has been stopped hence it is giving transport endpoint not connected error.
3. Adding has_glusterd_started check in the d_parent_test

PR #472, opened 2021-05-27
Mount cleanup corner case handling

srijan-sivakumar:
The mount cleanup is not done properly due to some corner cases
1. When the mountpount is already unmounted
2. When the mountpoint is unmount and deleted

And this is done in such a manner that the update escapes the data structure itself.

Fixes: #467


nishith-vihar (approved):
LGTM

aujjwal-redhat (approved):
LGTM

Issue #478, opened 2021-05-27
Node reboot handling


srijan-sivakumar:
Need operational functions to handle,

- [ ] Rebooting a node
- [ ] Waiting for it to come online
- [ ] To check if a node is online or offline [ Power state ]

PR #489, opened 2021-05-28
[FRMV] Machine ops and related test

srijan-sivakumar:
Added machine ops which takes care of machine
state. To support it, test has been added in example.

Also, enhanced the terminate of d_parent and nd_parent
tests.

Fixes: #478

### DIFF
--- a/common/ops/support_ops/machine_ops.py
+++ b/common/ops/support_ops/machine_ops.py
@@ -1,0 +1,111 @@
+"""
+Machine ops is the swiss army knife of operational functions required
+to perform server related configuration changes, be it network stack,
+systemd changes or maybe a node reboot itself.
+"""
+import time
+import os
+from common.ops.abstract_ops import AbstractOps
+
+
+class MachineOps(AbstractOps):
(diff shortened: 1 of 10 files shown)


common/ops/support_ops/machine_ops.py
```diff
@@ -0,0 +1,111 @@
...
+    MachineOps class provides methods for
+    handling the machine specific operations.
+    """
+
+    def reboot_nodes(self, nodes: list):
+        """
+        To reboot a given set of node(s)
+        Arg:
+            node(s) (str/list)
+        Returns:
```
aujjwal-redhat, line 22: It should be boolean

aujjwal-redhat (changes requested):

common/ops/support_ops/machine_ops.py
srijan-sivakumar: Ahh..this reminds me..I actually have to remove the return value from here :p

nik-redhat (approved):

Issue #496, opened 2021-05-28
[Common] get_bricks_online and get_bricks_offline function


srijan-sivakumar:
Following functions are required for test cases,

- [x] `get_bricks_offline`
- [x] `get_bricks_online`

PR #498, opened 2021-05-28
Brick ops enhancement

srijan-sivakumar:
Added the following functions
    1. get_bricks_offline
    2. get_bricks_online
    3. wait_for_bricks_to_go_offline
    4. wait_for_bricks_to_come_online
    
Fixes: #496

### DIFF
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -553,6 +553,81 @@ class BrickOps:
         self.logger.error(f"Key 'node' not in volume status of {volname}")
         return None
 
+    def wait_for_bricks_to_go_offline(self, volname: str, brick_list: list,
+                                      timeout: int = 100) -> bool:
+        """
+        Function to wait till the given set of bricks in brick list go offline.
+
+        Args:
+            volname (str): Name of the volume whose bricks are to be noticed.
+            brick_list (list): List of bricks which are to be brought down.
+            timeout (int): Optional parameter with defailt value 100. The
+            function waits for these many secondsat max till bricks go offline.
+
+        Returns:
+            True if the bricks go offline or False.
+        """
+        nd_list = []
+        for brickd in brick_list:
+            node, _ = brickd.split(':')
+            if node not in nd_list:
+                nd_list.append(node)
+        itr = 0
+        while itr < timeout:
+            random_node = random.choice(nd_list)
+            offline_brick_list = self.get_offline_bricks_list(volname,
+                                                              random_node)
+            if set(brick_list).issubset(set(offline_brick_list)):
+                return True
+            itr += 5
+            sleep(5)
+
+        offline_brick_list = self.get_offline_bricks_list(volname, nd_list[0])
+        if set(brick_list).issubset(set(offline_brick_list)):
+            return True
+        self.logger.error(f"Current offline brick list : {offline_brick_list}"
+                          " Compared to expected offline brick list :"
+                          f" {brick_list}")
+        return False
+
+    def wait_for_bricks_to_come_online(self, volname: str, server_list: list,
+                                       brick_list: list,
+                                       timeout: int = 100) -> bool:
+        """
+        Function to wait till the given set of bricks in brick list come
+        online.
+
+        Args:
+            volname (str): Name of the volume whose bricks are to be brought
+                           online.
+            server_list (list): A list of servers which are hosting the volume.
+            brick_list (list): List of bricks which are to be brought up.
+            timeout (int): Optional parameter with defailt value 100. The
+                           function waits for these many seconds at max till
+                           bricks come online.
+
+        Returns:
+            True if the bricks come online or False.
+        """
+        itr = 0
+        while itr < timeout:
+            random_node = random.choice(server_list)
+            online_brick_list = self.get_online_bricks_list(volname,
+                                                            random_node)
+            if set(brick_list).issubset(set(online_brick_list)):
+                return True
+            itr += 5
+            sleep(5)
+
+        online_brick_list = self.get_online_bricks_list(volname,
+                                                        server_list[0])
+        if set(brick_list).issubset(set(online_brick_list)):
+            return True
+        self.logger.error(f"Current online brick list : {online_brick_list}"
+                          " Compared to expected online brick list :"
+                          f" {brick_list}")
+        return False
+
     # TODO Brick mux logic inclusion.
     def bring_bricks_offline(self, volname: str, brick_list: list,
                              timeout: int = 100) -> bool:
@@ -563,6 +638,8 @@ class BrickOps:
             volname (str): Name of the volume whose bricks are to be brought
                            down.
             brick_list (list): List of bricks which are to be brought down.
+            timeout (int): Optional parameter with defailt value 100. The
+            function waits for these many secondsat max till bricks go offline.
 
         Returns:
             True if the bricks are brought offline or False.
@@ -583,23 +660,7 @@ class BrickOps:
                 return False
 
         # Wait till the said bricks come offline.
-        itr = 0
-        while itr < timeout:
-            random_node = random.choice(nd_list)
-            offline_brick_list = self.get_offline_bricks_list(volname,
-                                                              random_node)
-            if set(brick_list).issubset(set(offline_brick_list)):
-                return True
-            itr += 5
-            sleep(5)
-
-        offline_brick_list = self.get_offline_bricks_list(volname, nd_list[0])
-        if set(brick_list).issubset(set(offline_brick_list)):
-            return True
-        self.logger.error(f"Current offline brick list : {offline_brick_list}"
-                          " Compared to expected offline brick list :"
-                          f" {brick_list}")
-        return False
+        return self.wait_for_bricks_to_go_offline(volname, brick_list, timeout)
 
     # TODO Brick mux logic inclusion.
     def bring_bricks_online(self, volname: str, server_list: str,
@@ -642,21 +703,5 @@ class BrickOps:
         self.wait_till_all_peers_connected(server_list)
 
         # Wait till all said bricks are online.
-        itr = 0
-        while itr < timeout:
-            random_node = random.choice(server_list)
-            online_brick_list = self.get_online_bricks_list(volname,
-                                                            random_node)
-            if set(brick_list).issubset(set(online_brick_list)):
-                return True
-            itr += 5
-            sleep(5)
-
-        online_brick_list = self.get_online_bricks_list(volname,
-                                                        server_list[0])
-        if set(brick_list).issubset(set(online_brick_list)):
-            return True
-        self.logger.error(f"Current online brick list : {online_brick_list}"
-                          " Compared to expected online brick list :"
-                          f" {brick_list}")
-        return False
+        return self.wait_for_bricks_to_come_online(volname, server_list,
+                                                   brick_list, timeout)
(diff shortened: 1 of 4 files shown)


common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+            timeout (int): The timout till which we will check whether the
+                           bricks have come online.
+
+        Returns:
+            True if the bricks are online or False.
+        """
+        if not disrup_method:
+            self.logger.info(f"Getting bricks {brick_list} online by forced"
+                             f" volume start of {volname}")
+            self.volume_start(volname, random.choice(server_list), True)
```
nik-redhat, line 691: This will again be affected if glusterd is stopped or node rebooted

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+        else:
+            self.logger.info(f"Getting bricks {brick_list} online by "
+                             " glusterd restart in respective nodes.")
+            node_list = []
+            for bdata in brick_list:
+                if bdata.split(':')[0] not in node_list:
```
nik-redhat, line 697: We can use this logic at above function as well :p .ie, `wait_for_bricks_to_go_offline`

tests/example/sample_component/test_brick_ops.py
```diff
@@ -29,8 +28,25 @@ def run_test(self, redant):
                                             self.server_list[0])
         redant.logger.info(f"Online bricks:{ret}")
 
+        brick_list = ret[0:1]
```
nik-redhat, line 31: We can simply use `ret[0]` as well?

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+        Args:
+            volname (str): Name of the volume whose bricks are to be brought
+                           down.
+            brick_list (list): List of bricks which are to be brought down.
+            timeout (int): Optional parameter with defailt value 100. The
+            function waits for these many secondsat max till bricks go offline.
+
+        Returns:
+            True if the bricks are brought offline or False.
+        """
```
nik-redhat, line 646: A check for `brick_list` not being a list is needed, and converting it to list

tests/example/sample_component/test_brick_ops.py
```diff
@@ -29,8 +28,25 @@ def run_test(self, redant):
...
+        ret = redant.bring_bricks_online(self.vol_name, self.server_list,
+                                         brick_list)
+        if not ret:
+            raise Exception(f"{brick_list} still offline.")
+
         redant.stop_glusterd(self.server_list[1])
-        sleep(2)
+        redant.wait_for_glusterd_to_stop(self.server_list[1])
+        if redant.is_glusterd_running(self.server_list[1]):
+            raise Exception(f"Glusterd not stopped on {self.server_list[1]}")
```
nik-redhat, line 48: The above function also makes a call to `is_glusterd_running` internally, so it doesn't make sense to use it here again.

tests/example/sample_component/test_brick_ops.py
```diff
@@ -42,9 +58,12 @@ def run_test(self, redant):
         redant.logger.info(f"Offline bricks:{ret}")
 
         redant.start_glusterd(self.server_list[1])
-        redant.logger.info(f"Started glusterd on {self.server_list[1]}")
 
-        sleep(2)
+        redant.wait_for_glusterd_to_start(self.server_list[1])
+        if not redant.is_glusterd_running(self.server_list[1]):
+            raise Exception(f"Glusterd not started on {self.server_list[1]}")
```
nik-redhat, line 64: Same as above

tests/example/sample_component/test_brick_ops.py
nik-redhat: Instead, simply check the ret value of above function call

nik-redhat (commented):
I am not able to understand the subset logic, how is it working and why is it needed?

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+    def wait_for_bricks_to_go_offline(self, volname: str, brick_list: list,
+                                      timeout: int = 100) -> bool:
+        """
+        Function to wait till the given set of bricks in brick list go offline.
+
+        Args:
+            volname (str): Name of the volume whose bricks are to be noticed.
+            brick_list (list): List of bricks which are to be brought down.
+            timeout (int): Optional parameter with defailt value 100. The
+            function waits for these many secondsat max till bricks go offline.
```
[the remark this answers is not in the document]
srijan-sivakumar: so, with the subset, what we are doing is checking if the list which we have ( expectation ) is present in the actual list. Technically it is like saying check if (1, 2, 4) is a subset of (1, 2, 3, 4, 5).

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+
+        Returns:
+            True if the bricks come online or False.
+        """
+        itr = 0
+        while itr < timeout:
+            random_node = random.choice(server_list)
+            online_brick_list = self.get_online_bricks_list(volname,
+                                                            random_node)
+            if set(brick_list).issubset(set(online_brick_list)):
```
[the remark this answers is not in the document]
srijan-sivakumar: Same as above. in comment 1

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+        for brick in brick_list:
+            nd, path = brick.split(":")
+            if nd not in nd_list:
+                nd_list.append(nd)
+            path = path.replace("/", "-")
+            cmd = (f"pid=`ps -ef | grep -ve 'grep' | grep -e '{nd}{path}.pid' "
+                   " | awk '{print $2}'` && kill -15 $pid || kill -9 $pid")
+            ret = self.execute_abstract_op_node(cmd, nd, False)
+            if ret['error_code'] != 0:
+                self.logger.error(f"Failed to get {brick_list} offline."
```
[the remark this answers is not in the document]
srijan-sivakumar: ack

tests/example/sample_component/test_brick_ops.py
srijan-sivakumar: actually, ret[0] will return a string or value at ret[0] but what I wanted was a list, hence ret[0:1]. So, either we can represent it as ret[0:1] or [ret[0]]

common/ops/gluster_ops/brick_ops.py
srijan-sivakumar: Ack.

common/ops/gluster_ops/brick_ops.py
```diff
@@ -556,3 +552,156 @@ def get_offline_bricks_list(self, volname: str, node: str) -> list:
...
+            timeout (int): Optional parameter with defailt value 100. The
+                           function waits for these many seconds at max till
+                           bricks come online.
+
+        Returns:
+            True if the bricks come online or False.
+        """
+        itr = 0
+        while itr < timeout:
+            random_node = random.choice(server_list)
```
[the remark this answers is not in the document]
srijan-sivakumar: well..in such cases, we should rather use glusterd_start before waiting for bricks to come online. In the second scenario again, we should be waiting for the node to come online before calling these functions. But if the user wants to see a test case wherein they just wants to do something as remote as bringing glusterd to stop and then waiting for the nodes to come up, then we can let them do that and let them handle the failure ( which is expected ).

common/ops/gluster_ops/brick_ops.py
srijan-sivakumar: Ack

common/ops/gluster_ops/brick_ops.py
srijan-sivakumar: Ack.

tests/example/sample_component/test_brick_ops.py
srijan-sivakumar: Ack...I was experimenting with some other operation. Will remove these checks.

tests/example/sample_component/test_brick_ops.py
srijan-sivakumar: Ack.

common/ops/gluster_ops/brick_ops.py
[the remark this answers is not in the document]
nik-redhat: So, why do a check again at line: 585 again?

common/ops/gluster_ops/brick_ops.py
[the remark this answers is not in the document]
nik-redhat: My point is why do random selection? And not allow the user to pass the node on which the command is to be run?

tests/example/sample_component/test_brick_ops.py
nik-redhat: Ohh, got your point......but better to have a check in the ops function, so that either way works.

common/ops/gluster_ops/brick_ops.py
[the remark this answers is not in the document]
srijan-sivakumar: Now..it might be possible that we keep requesting the same info for sometime.. (maybe user gave en extended timeout ). Now,  instead of forcing all requests to a single server, we can distribute it across to other servers too.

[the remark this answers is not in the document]
nik-redhat: Ok....makes sense.

common/ops/gluster_ops/brick_ops.py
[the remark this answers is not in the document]
srijan-sivakumar: That was just a final check before we exit...technically we sleep for 5 seconds and say this time, we exceed the limit of timeout, then it might be possible that in those 5 seconds of sleep, the bricks might've come online. Now you might ask, why not keep the check after sleep inside the while loop itself. We can but suppose the bricks went offline/online ( this logic is present in both functions) quickly then why to waste 5 seconds on sleep. It is an extra line..but it can save us some pain if things work as above.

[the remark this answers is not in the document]
nik-redhat: See if you consider that scenario, there won't be an end to possibilities. It might be possible that the bricks come online just after the last execution, so shall we check again? No, of course. So, better to not provide extra calculations which don't guarantee many benefits. 20secs is a lot of time for a brick to come online, if it is taking ore than that then the system is not capable of hosting the bricks or running the framework.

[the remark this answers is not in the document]
nik-redhat: Even if you are worried we might face such issues, increase the time a bit, i.e make 100sec to 120secs.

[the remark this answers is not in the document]
srijan-sivakumar: All right :p Then I'll remove the final checks.

nik-redhat (approved):

nik-redhat:
Could you add the test output as well? Just for confirmation...

srijan-sivakumar:
```(redant-demo) [root@rhsdev-virt05 redant]# python3 core/redant_main.py -c ../rconfig.yml -t tests/example/sample_component/test_brick_ops.py -ll D -sp
test_brick_ops-dist-rep PASS

Summary:
Table:
 test_brick_ops
 
|   dist-rep  |     PASS    | 47.549620389938354 |



Framework runtime : 51.876975774765015```

PR #565, opened 2021-06-01
Update volume_ops.py

srijan-sivakumar:
Bypassing the force option in create_volume as the requirement is to setup the volume at any cost. Because gluster is snobby about creating bricks on root dir.


aujjwal-redhat (approved):
LGTM

nik-redhat (approved):

PR #567, opened 2021-06-01
[Test] Making reset optional

srijan-sivakumar:
Resetting the volume options seems to be a costly transaction. Hence making it optional.

### DIFF
--- a/tests/d_parent_test.py
+++ b/tests/d_parent_test.py
@@ -34,6 +34,7 @@ class DParentTest(metaclass=abc.ABCMeta):
         server_details = param_obj.get_server_config()
         client_details = param_obj.get_client_config()
 
+        self.reset_option = False
         self.TEST_RES = True
         self.volume_type = volume_type
         self.vol_type_inf = param_obj.get_volume_types()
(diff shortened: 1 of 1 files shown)


tests/d_parent_test.py
```diff
@@ -110,7 +111,9 @@ def terminate(self):
             self.redant.logger.error(tb)
 
         try:
-            self.redant.reset_volume_option('all', 'all', self.server_list[0])
+            if self.reset_option:
```
nik-redhat, line 114: Aree....but the TCs need to be updated for this :thinking: , to actually allow cluster reset.

srijan-sivakumar: Yup..but I see not all cases needs a option reset. And this seems to be the root of failure of the env itself yesterday...

srijan-sivakumar: We will take this as something to be done by user in a test case. That way..we can prevent a costly operation. With time, we can replace this with a volds based modification.

nik-redhat: Yeah, that's what I am saying :sweat_smile: , so now if some cases are setting cluster level option it won't be reset, if we do so and not update the TCs in the same patch

srijan-sivakumar: Will take that as a next patch. I do see 2-3 cases setting cluster level options. But most are setting only volume level options.

nik-redhat (approved):

Issue #573, opened 2021-06-02
[TestBug] Test Peer Detach


srijan-sivakumar:
Couple of cases which have to be considered in test_peer_detach

1. The detach of the node which has already been detached comes out to be true. But on a finer look into the logs, it shows that it had actually failed as expected.

```
[2021-06-02 04:26:27,486] INFO  - Validation after peer probe is successful.
[2021-06-02 04:26:27,487] ERROR  - 'TestCase' object has no attribute 'volume_name'
[2021-06-02 04:26:27,487] ERROR  - Traceback (most recent call last):
  File "./tests/functional/glusterd/test_peer_detach.py", line 40, in terminate
    self.redant.cleanup_volume(self.volume_name, self.server_list[0])
AttributeError: 'TestCase' object has no attribute 'volume_name'
{'msg': {'opRet': '-1', 'opErrno': '2', 'opErrstr': '10.1.41.240 is not part of cluster'}, 'Flag': True, 'node': '10.1.41.34', 'cmd': 'gluster --xml peer detach 10.1.41.240 --mode=script', 'error_code': 0
```

2. Secondly, the volume creation comes after the detach stage. If something had happened before the volume_Creation, we shouldn't be performing a cleanup of the same. Or rather the check condition should be modified in such a manner that the vol_name variable is created at the very beginning.

PR #577, opened 2021-06-02
Fix: Peer probe and detach TCs to check for 'opRet' value

nik-redhat:
Fix:
Peer probe and detach operation updates the `opRet` and `opErrstr` value instead of
the `error_code` and `error_msg`, so updated the Tcs and the ops accordingly.

Fixes: #573 

### DIFF
--- a/common/ops/gluster_ops/peer_ops.py
+++ b/common/ops/gluster_ops/peer_ops.py
@@ -79,7 +79,8 @@ class PeerOps(AbstractOps):
                 if 'output' not in ret['msg']:
                     return False
                 if (ret['error_code'] != 0
-                        or ret['msg']['output'] != 'success'):
+                        or ret['msg']['output'] != 'success'
+                        or ret['msg']['opRet'] != '0'):
                     self.logger.error("Failed to peer probe the node"
                                       f" {server}")
                     return False
@@ -167,7 +168,8 @@ class PeerOps(AbstractOps):
         for server in servers:
             ret = self.peer_detach(server, node, force, False)
             if (ret['error_code'] != 0
-                    or ret['msg']['output'] != 'success'):
+                    or ret['msg']['output'] != 'success'
+                    or ret['msg']['opRet'] != '0'):
                 self.logger.error(f"Failed to detach the node {server}")
                 return False
 
--- a/tests/functional/glusterd/test_peer_detach.py
+++ b/tests/functional/glusterd/test_peer_detach.py
@@ -37,7 +37,11 @@ class TestCase(DParentTest):
                                                  self.server_list[0])
             if not ret:
                 raise Exception("Peer probing failed on some of the servers")
-            self.redant.cleanup_volume(self.volume_name, self.server_list[0])
+
+            # Cleanup volume, if created
+            if self.is_vol_created:
+                self.redant.cleanup_volume(self.volume_name,
+                                           self.server_list[0])
 
         except Exception as error:
             tb = traceback.format_exc()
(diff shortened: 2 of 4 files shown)


nik-redhat:
`test_probe_hostname` should still be under excluded list as it has issues with fqdn probing.

nik-redhat:
Test run:
```
Table:
 test_peer_detach
 
|   Generic   |     PASS    | 57.02109479904175 |

Framework runtime : 59.03855848312378
```

```
Table:
 test_peer_status
 
|   Generic   |     PASS    | 32.72476887702942 |

Framework runtime : 34.665186405181885
```

tests/functional/glusterd/test_peer_detach.py
```diff
@@ -75,6 +79,7 @@ def run_test(self, redant):
         - Peer detach one node which hosts bricks of offline volume
         - Peer detach force a node which hosts bricks of offline volume
         """
+        self.is_vol_created = False
```
nishith-vihar, line 82: I think this check might not be needed as cleanup volume anyways checks for volume existance in volds.

nik-redhat:
Test run:
```
Table:
 test_peer_detach
 
|   Generic   |     PASS    | 39.06959939002991 |



Framework runtime : 41.08415865898132
```

nishith-vihar (approved):
LGTM

tests/functional/glusterd/test_peer_detach.py
```diff
@@ -106,14 +86,14 @@ def run_test(self, redant):
...
                                  False, False)
-        if ret['error_code'] == 0:
+        if ret['msg']['opRet'] == '0':
             raise Exception(f"Server detach invalid host should have "
                             f"failed: {self.invalid_ip}")
 
         # Detach non exist host
         ret = redant.peer_detach(self.non_exist_host, self.server_list[0],
                                  False, False)
-        if ret['error_code'] == 0:
```
srijan-sivakumar, line 116: if peer_detach fails, the rexe will not get any 'msg' value and hence this might throw a keyerror when accessing from ret['msg'['opRet']. What we can do is to check the error code and based on that take a decision whether to access ret['msg']

srijan-sivakumar (changes requested):

aujjwal-redhat (approved):
LGTM

tests/functional/glusterd/test_peer_detach.py
nik-redhat: Peer detach, do fail here and we get a `msg` dictionary in return :thinking:

common/ops/gluster_ops/peer_ops.py
```diff
@@ -79,7 +79,8 @@ def peer_probe_servers(self, servers: list, node: str,
                 if 'output' not in ret['msg']:
                     return False
                 if (ret['error_code'] != 0
-                        or ret['msg']['output'] != 'success'):
+                        or ret['msg']['output'] != 'success'
```
srijan-sivakumar, line 82: Now...if `ret['error_code']` is not zero, there won't be any `ret['msg']` value as rexe doesn't add message if it is a failure. Hence we need `and` here after `ret['error_code']` or else we would see the second condition being checked and we would see the failure as 'msg' doesn't exists as a key for ret.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_peer_detach.py
srijan-sivakumar: but still it would be better to keep a check for error code also. Now I think in this scenario the error code might be 0, hence the msg would be populated. But there are scenarios when error_code won't be 0 and then we will see a failure. There can be many scenarios..if there is some timeout, we would be getting error_code as non zero hence msg won't be populated.

common/ops/gluster_ops/peer_ops.py
nik-redhat: We do add a `msg` in the ret even if `error_code` is not 0, as per this code structure: https://github.com/srijan-sivakumar/redant/blob/main/common/rexe.py#L117-L134

tests/functional/glusterd/test_peer_detach.py
nik-redhat: We are doing a `-- xml` execution, so I think `opRet` will always be there is `msg`.

common/ops/gluster_ops/peer_ops.py
srijan-sivakumar: 🤔 ok...cool. Then its fine. last I remembered, this wasn't present. Anyway if that's the case then we are good to go.

srijan-sivakumar (approved):

Issue #575, opened 2021-06-02
[Test] Tasks key-error


srijan-sivakumar:
The test giving trouble being : test_rebalance_when_quorum_not_met

```[2021-06-02 04:39:27,595] INFO  - Running gluster volume status test_rebalance_when_quorum_not_met-dist   --xml on 10.1.41.34
[2021-06-02 04:39:27,723] DEBUG  - {'msg': {'opRet': '0', 'opErrno': '0', 'opErrstr': None, 'volStatus': {'volumes': {'volume': {'volName': 'test_rebalance_when_quorum_not_met-dist', 'nodeCount': '3', 'node': [{'hostname': '10.1.41.34', 'path': '/home/glusterfs/bricks/test_rebalance_when_quorum_not_met-dist-0', 'peerid': '99c85768-7f5a-45b6-a98f-534dcb1d73c8', 'status': '0', 'port': 'N/A', 'ports': {'tcp': 'N/A', 'rdma': 'N/A'}, 'pid': '-1'}, {'hostname': '10.1.41.240', 'path': '/home/glusterfs/bricks/test_rebalance_when_quorum_not_met-dist-1', 'peerid': '536c4a21-d80b-448e-9870-ac11e1f602af', 'status': '0', 'port': 'N/A', 'ports': {'tcp': 'N/A', 'rdma': 'N/A'}, 'pid': '-1'}, {'hostname': '10.1.41.51', 'path': '/home/glusterfs/bricks/test_rebalance_when_quorum_not_met-dist-2', 'peerid': '10c0d179-410c-408b-9b48-024ac362e4ff', 'status': '0', 'port': 'N/A', 'ports': {'tcp': 'N/A', 'rdma': 'N/A'}, 'pid': '-1'}], 'tasks': None}}}}, 'Flag': True, 'node': '10.1.41.34', 'cmd': 'gluster volume status test_rebalance_when_quorum_not_met-dist   --xml', 'error_code': 0}
[2021-06-02 04:39:27,724] ERROR  - 'tasks'
[2021-06-02 04:39:27,724] ERROR  - Traceback (most recent call last):
  File "./tests/d_parent_test.py", line 88, in parent_run_test
    self.run_test(self.redant)
  File "./tests/functional/glusterd/test_rebalance_when_quorum_not_met.py", line 75, in run_test
    if ret[self.vol_name]['tasks'] is not None:
KeyError: 'tasks'```

PR #579, opened 2021-06-02
[Testfix] Changing the key value

srijan-sivakumar:
The key value is `task_status` and not `task`.

Fixes: #575


nishith-vihar (approved):
LGTM

PR #580, opened 2021-06-02
Removing test_probe_hostname from exclude list

srijan-sivakumar:
With recent fixes, we can run test_probe_hostname. Hence removing it from the exclude list.

Signed-off-by: srijan-sivakumar ssivakum@redhat.com


srijan-sivakumar:
Test is still causing issues.

Issue #562, opened 2021-06-01
[TEST] Migration of test_volume_create.py


nishith-vihar:
Migration of [test_case](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_volume_create.py) from glusto to redant

PR #583, opened 2021-06-02
Migration of test_volume_create.py

nishith-vihar:
Migration of [test_volume_create.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_volume_create.py) from glusto to redant

Updates: #292
Fixes: #562 

### DIFF
--- a/tests/functional/glusterd/test_volume_create.py
+++ b/tests/functional/glusterd/test_volume_create.py
@@ -41,7 +41,7 @@ class TestVolumeCreate(GlusterBaseClass):
     def setUpClass(cls):
 
         # Calling GlusterBaseClass setUpClass
-        GlusterBaseClass.setUpClass.im_func(cls)
+        cls.get_super_method(cls, 'setUpClass')()
 
         ret = cls.validate_peers_are_connected()
         if not ret:
@@ -73,7 +73,7 @@ class TestVolumeCreate(GlusterBaseClass):
                 raise ExecutionError("Unable to delete volume % s" % volume)
             g.log.info("Volume deleted successfully : %s", volume)
 
-        GlusterBaseClass.tearDown.im_func(self)
+        self.get_super_method(self, 'tearDown')()
 
     def test_volume_create(self):
         '''
--- a/tests/functional/glusterd/test_volume_create.py
+++ b/tests/functional/glusterd/test_volume_create.py
@@ -23,7 +23,7 @@ import random
 from tests.d_parent_test import DParentTest
 
 
-# disruptive;dist
+# disruptive;
 class TestCase(DParentTest):
 
     def run_test(self, redant):
@@ -62,8 +62,8 @@ class TestCase(DParentTest):
         if not ret:
             raise Exception("Failed to bring down the bricks")
 
-        ret = redant.volume_start(self.volume_name1, self.server_list[0],
-                                  force=True)
+        redant.volume_start(self.volume_name1, self.server_list[0],
+                            force=True)
 
         ret = redant.wait_for_bricks_to_come_online(self.volume_name1,
                                                     self.server_list,
@@ -75,27 +75,34 @@ class TestCase(DParentTest):
         self.volume_type2 = 'dist'
         self.volume_name2 = f"{self.test_name}-{self.volume_type2}-2"
         conf_dict = self.vol_type_inf[self.conv_dict[self.volume_type2]]
+        excep = True
         try:
             redant.volume_create_with_custom_bricks(self.volume_name2,
                                                     self.server_list[0],
                                                     conf_dict, brick_cmd,
                                                     brick_dict)
-            raise Exception("Unexpected: Successfully created the volume "
-                            "with previously used bricks")
+            excep = False
         except Exception:
             redant.logger.info("Expected: Failed to create a volume "
                                "with previously used bricks.")
+
+        if not excep:
+            raise Exception("Unexpected: Successfully created the volume "
+                            "with previously used bricks")
 
         # create a volume with already existing volume name
         try:
             redant.setup_volume(self.volume_name1, self.server_list[0],
                                 conf_dict, self.server_list,
-                                self.brick_roots, True)
-            raise Exception("Unexpected: Successfully created the volume with "
-                            "already existing volname")
+                                self.brick_roots)
+            excep = False
         except Exception:
             redant.logger.info("Expected: Failed to create the volume with "
                                "already existing volname")
+
+        if not excep:
+            raise Exception("Unexpected: Successfully created the volume with "
+                            "already existing volname")
 
         # creating a volume with non existing brick path should fail
         self.volume_type3 = 'dist'
@@ -113,11 +120,14 @@ class TestCase(DParentTest):
                                                     self.server_list[0],
                                                     conf_dict, brick_cmd,
                                                     brick_dict)
-            raise Exception("Unexpected: Successfully created the volume "
-                            "with non existing brick path")
+            excep = False
         except Exception:
             redant.logger.info("Expected: Failed to create the volume with "
                                "non existing brick path")
+
+        if not excep:
+            raise Exception("Unexpected: Successfully created the volume "
+                            "with non existing brick path")
 
         # cleanup the volume and peer detach all servers. form two clusters,try
         # to create a volume with bricks whose nodes are in different clusters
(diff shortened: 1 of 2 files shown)


nishith-vihar:
Required 4 servers in the test case logic

nik-redhat:
Rebase the PR, and add a check at the top for the servers and let's take this PR to completion.

common/ops/gluster_ops/volume_ops.py
```diff
@@ -69,7 +69,8 @@ def setup_volume(self, volname: str, node: str, conf_hash: dict,
 
     def volume_create(self, volname: str, node: str, conf_hash: dict,
```
nik-redhat, line 70: A new function for this is added, `volume_create_with_custome_brick` or something of that sort, use that.

nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,175 @@
...
+        redant.logger.info("Get all the bricks of the volume")
+        bricks_list = redant.es.get_all_bricks_list(self.volume_name1)
+        if len(bricks_list) == 0:
+            raise Exception("Failed to get the brick list")
+
+        ret = redant.bring_bricks_offline(self.volume_name1, bricks_list[0:2])
+        if not ret:
+            raise Exception("Failed to bring down the bricks")
+
+        ret = redant.volume_start(self.volume_name1, self.server_list[0],
```
nik-redhat, line 65: ret not needed

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+        # create volume with previously used bricks and different volume name
+        self.volume_type2 = 'dist'
+        self.volume_name2 = f"{self.test_name}-{self.volume_type2}-2"
+        conf_dict = self.vol_type_inf[self.conv_dict[self.volume_type2]]
+        try:
+            redant.volume_create_with_custom_bricks(self.volume_name2,
+                                                    self.server_list[0],
+                                                    conf_dict, brick_cmd,
+                                                    brick_dict)
+            raise Exception("Unexpected: Successfully created the volume "
```
nik-redhat, line 83: You are raising an exception here and handling it the next line by saying `expected` which is not true.
Instead, use the excep flag and check the error_code in the return value and raise an exception accordingly.

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+        except Exception:
+            redant.logger.info("Expected: Failed to create a volume "
+                               "with previously used bricks.")
+
+        # create a volume with already existing volume name
+        try:
+            redant.setup_volume(self.volume_name1, self.server_list[0],
+                                conf_dict, self.server_list,
+                                self.brick_roots, True)
+            raise Exception("Unexpected: Successfully created the volume with "
```
nik-redhat, line 94: Same as above

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+  Test glusterd behavior with the gluster volume create command
+"""
+
+import random
+from tests.d_parent_test import DParentTest
+
+
+# disruptive;dist
+class TestCase(DParentTest):
+
```
nik-redhat, line 28: Use the setup_test function to override the volume creation in the parent_run_test.

nik-redhat: Also, we can simply make it `disruptive`, we don't need to create a dist volume in the parent_run_test

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+                            "with previously used bricks")
+        except Exception:
+            redant.logger.info("Expected: Failed to create a volume "
+                               "with previously used bricks.")
+
+        # create a volume with already existing volume name
+        try:
+            redant.setup_volume(self.volume_name1, self.server_list[0],
+                                conf_dict, self.server_list,
+                                self.brick_roots, True)
```
nik-redhat, line 93: No need for `force`, also, pass `False` to except to get the return value accordingly.

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+        try:
+            redant.volume_create_with_custom_bricks(self.volume_name3,
+                                                    self.server_list[0],
+                                                    conf_dict, brick_cmd,
+                                                    brick_dict)
+            raise Exception("Unexpected: Successfully created the volume "
+                            "with non existing brick path")
+        except Exception:
+            redant.logger.info("Expected: Failed to create the volume with "
+                               "non existing brick path")
```
nik-redhat, line 120: Same as above

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+        # cluster should fail
+        try:
+            redant.setup_volume(self.volume_name3, self.server_list[0],
+                                conf_dict, self.server_list,
+                                self.brick_roots, True)
+            raise Exception("Unexpected: Successfully created the volume "
+                            "with bricks which are part of another cluster")
+        except Exception:
+            redant.logger.info("Expected: Failed to create the volume with "
+                               "bricks which are part of another cluster")
```
nik-redhat, line 152: Same as above.
No need for `force` and use except flag for return value

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,178 @@
...
+        conf_dict = self.vol_type_inf[self.conv_dict[self.volume_type4]]
+        try:
+            redant.setup_volume(self.volume_name4, self.server_list[0],
+                                conf_dict, self.server_list,
+                                self.brick_roots, True)
+            raise Exception("Unexpected: Successfully created the volume "
+                            "with brick whose node is down")
+        except Exception:
+            redant.logger.info("Expected: Failed to create the volume with "
+                               "brick whose node is down")
```
nik-redhat, line 178: Same as above

nik-redhat (changes requested):

tests/functional/glusterd/test_volume_create.py
nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
nishith-vihar: Ack

tests/functional/glusterd/test_volume_create.py
nishith-vihar: I don't think we need to set excep flag to False as we are not using the ret value anywhere

tests/functional/glusterd/test_volume_create.py
nishith-vihar: I dont think we would need to set the excep flag to False here also. same as above

tests/functional/glusterd/test_volume_create.py
nishith-vihar: I think it would be just enough to remove the dist volume just make it disruptive

nik-redhat: Yes, that would do as well.

tests/functional/glusterd/test_volume_create.py
nik-redhat: The only reason to use `False` is that if we except the Exception here and just log it then we will have to logs now at the ERROR level done by the abstract ops and another here at INFO level saying it is expected which will create disparities.

tests/functional/glusterd/test_volume_create.py
```diff
@@ -0,0 +1,194 @@
...
+        self.volume_type2 = 'dist'
+        self.volume_name2 = f"{self.test_name}-{self.volume_type2}-2"
+        conf_dict = self.vol_type_inf[self.conv_dict[self.volume_type2]]
+        excep = True
+        try:
+            redant.volume_create_with_custom_bricks(self.volume_name2,
+                                                    self.server_list[0],
+                                                    conf_dict, brick_cmd,
+                                                    brick_dict)
+            excep = False
```
nik-redhat, line 84: Hey this is not the correct way of using it, we need to pass it in the function itself, check the function definition for the usage.

nik-redhat (approved):

nik-redhat (approved):

PR #587, opened 2021-06-02
Handle DS inconsistency in volume ops

srijan-sivakumar:
During TC failures, the data structure might be inconsistent hence bypassing the excep check.


nik-redhat:
Is this really needed after the last update to the ops?

aujjwal-redhat (approved):
LGTM

srijan-sivakumar:
> Is this really needed after the last update to the ops?

Yes...if a TC fails..and if a volume is already started but the data structure wasn't updated..this is causing issues.

Issue #429, opened 2021-05-24
[TEST] Adding test_peer_probe_firewall_ports_not_opened


neesingh-rh:
Migrate the test case: [test_peer_probe_firewall_ports_not_opened](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py)

PR #631, opened 2021-06-05
Migrated TC test_peer_probe_firewall_ports_not_opened

nik-redhat:
Migrated TC [test_peer_probe_firewall_ports_not_opened](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py)

Fixes: #429 
Updates: #292 

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -57,3 +57,4 @@ excluded_tests:
     - tests/functional/glusterd/test_quorum_syslog.py
     - tests/functional/glusterd/test_volume_set_when_glusterd_stopped_on_one_node.py
     - tests/functional/glusterd/test_gluster_volume_status_xml_dump.py
+    - tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py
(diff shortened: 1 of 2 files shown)


nik-redhat:
This test is only applicable if we hace the setup done with firewall enabled and using rpm installation, because in case of rpm installation we get the firewall service files for glusterfs and rpm, is what I understand.

srijan-sivakumar:
> This test is only applicable if we hace the setup done with firewall enabled and using rpm installation, because in case of rpm installation we get the firewall service files for glusterfs and rpm, is what I understand.

The CI does have rpm installed env.

nik-redhat:
Ok, but is the peering done with firewall enabled? I don't think so..

srijan-sivakumar:
> Ok, but is the peering done with firewall enabled? I don't think so...

No..we still have to test that. I'm thinking of doing the testings post 9th...

nik-redhat:
Cool, marked under excluded, for the time being, we are good with this PR.

srijan-sivakumar (approved):

Issue #626, opened 2021-06-04
Migration of test_detach_node_used_to_mount.py


nishith-vihar:
Migration of [test_detach_node_used_to_mount.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_detach_node_used_to_mount.py) from glusto to redant

PR #633, opened 2021-06-07
Migration of test_detach_node_used_to_mount.py

nishith-vihar:
Migration of [test_detach_node_used_to_mount.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/glusterd/test_detach_node_used_to_mount.py) from glusto to redant.
get_fattr_list function is added in the io_ops which is needed in the test case.

Updates: #292
Fixes: #626

### DIFF
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -806,8 +806,8 @@ class IoOps(AbstractOps):
         """
         List of xattr for filepath on remote system.
         Args:
-            host (str): The hostname/ip of the remote system.
             fqpath (str): The fully-qualified path to the file.
+            node (str): The hostname/ip of the remote system.
         Kwargs:
         encode_hex(bool): Fetch xattr in hex if True
                             (Default:False)
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -801,6 +801,35 @@ class IoOps(AbstractOps):
         ret = self.execute_abstract_op_node(cmd, node)
         return ret['msg']
 
+    def get_fattr_list(self, fqpath: str, node: str,
+                       encode_hex=False) -> dict:
+        """
+        List of xattr for filepath on remote system.
+        Args:
+            host (str): The hostname/ip of the remote system.
+            fqpath (str): The fully-qualified path to the file.
+        Kwargs:
+        encode_hex(bool): Fetch xattr in hex if True
+                            (Default:False)
+        Returns:
+            xattr_list (dict): dict of xattr values.
+        """
+
+        cmd = f"getfattr --absolute-names -d -m - {fqpath}"
+        if encode_hex:
+            cmd = f"getfattr --absolute-names -d -m - -e hex {fqpath}"
+
+        ret = self.execute_abstract_op_node(cmd, node)
+
+        xattr_list = {}
+        for xattr_string in ret['msg'].strip().split('\n'):
+            xattr = xattr_string.split('=', 1)
+            if len(xattr) > 1:
+                key, value = xattr
+                xattr_list[key] = value
+
+        return xattr_list
+
     def set_fattr(self, fpath: str, fattr: str, node: str, value: str) -> list:
         """
         Function to set fattr on a path.
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -822,7 +822,8 @@ class IoOps(AbstractOps):
         ret = self.execute_abstract_op_node(cmd, node)
 
         xattr_list = {}
-        for xattr_string in ret['msg'].strip().split('\n'):
+        ret_val = "".join(ret['msg'])
+        for xattr_string in ret_val.strip().split('\n'):
             xattr = xattr_string.split('=', 1)
             if len(xattr) > 1:
                 key, value = xattr
--- a/tests/functional/glusterd/test_detach_node_used_to_mount.py
+++ b/tests/functional/glusterd/test_detach_node_used_to_mount.py
@@ -1,90 +1,50 @@
-#  Copyright (C) 2019-2020  Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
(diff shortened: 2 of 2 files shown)


nishith-vihar:
Requires 3 servers and 1 client for the test case run.

nik-redhat:
You can fix the lint, then I will check in my env and update the results.

common/ops/support_ops/io_ops.py
```diff
@@ -776,6 +776,35 @@ def get_fattr(self, fpath: str, fattr: str, node: str,
         ret = self.execute_abstract_op_node(cmd, node)
         return ret['msg']
 
+    def get_fattr_list(self, fqpath: str, node: str,
+                       encode_hex=False) -> dict:
+        """
+        List of xattr for filepath on remote system.
+        Args:
+            host (str): The hostname/ip of the remote system.
```
nik-redhat, line 784: nit: node

common/ops/support_ops/io_ops.py
```diff
@@ -776,6 +776,35 @@ def get_fattr(self, fpath: str, fattr: str, node: str,
...
+            fqpath (str): The fully-qualified path to the file.
```
nik-redhat, line 785: interchange the sequence of the args, i.e, first fqpath then node

nik-redhat:
The test requires 4 servers at least and 1 client which is not among that 4 servers.
Also, @nishith-vihar in this case you will have to mount the volume yourself and not use the predefined mount code, you can use the `setup_test` function for this. We need to mount the volume using the 4th server, and then detach that 4th server and do the operations. Check the glusto again for the logic.

tests/functional/glusterd/test_detach_node_used_to_mount.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+# disruptive;rep
+class TestCase(DParentTest):
+
+    def terminate(self):
+        try:
+            ret = (self.redant.
+                   wait_for_rebalance_to_complete(self.vol_name,
+                                                  self.server_list[0]))
```
nik-redhat, line 36: As per the TC, we don't have to wait for it to complete, rather simply stop it and continue.

tests/functional/glusterd/test_detach_node_used_to_mount.py
```diff
@@ -0,0 +1,140 @@
...
+        command = f"mkdir -p {self.mountpoint}/dir1"
+        redant.execute_abstract_op_node(command, self.client_list[0])
+
+        # Creating 100 files.
+        command = ('for number in `seq 101 200`;do touch '
+                   + self.mountpoint + '/file$number; done')
+        redant.execute_abstract_op_node(command, self.client_list[0])
+
+        # Forming brick list
+        brick_str = redant.form_brick_cmd(self.server_list, self.brick_roots,
```
nik-redhat, line 83: the form_brick_cmd function returns `dict, string` so update the return value accordingly.

tests/functional/glusterd/test_detach_node_used_to_mount.py
```diff
@@ -0,0 +1,140 @@
...
+                                          self.vol_name, 3, True)
+
+        # Adding bricks
+        redant.add_brick(self.vol_name, brick_str, self.server_list[0],
+                         replica_count=3)
+
+        # Start rebalance for volume.
+        redant.logger.info("Starting rebalance on the volume")
```
nik-redhat, line 91: logging not required

tests/functional/glusterd/test_detach_node_used_to_mount.py
```diff
@@ -0,0 +1,140 @@
...
+        redant.rebalance_start(self.vol_name, self.server_list[0])
+
+        # Creating 100 files.
+        command = ('for number in `seq 201 300`;do touch '
+                   + self.mounts[0].mountpoint + '/file$number; done')
+        redant.execute_abstract_op_node(command, self.client_list[0])
+
+        brick_list = redant.get_all_bricks(self.vol_name, self.server_list[0])
```
nik-redhat, line 99: We only need to check on the newly added bricks, don't fetch the complete brick list. Convert the brick_string received from the `form_brick_cmd` to a list nd use that.

nik-redhat (changes requested):

common/ops/support_ops/io_ops.py
nishith-vihar: Ack

common/ops/support_ops/io_ops.py
nishith-vihar: Ack

tests/functional/glusterd/test_detach_node_used_to_mount.py
nishith-vihar: This is in the terminate function right. Dont we need to wait for rebalance to complete if the test fails midway?

tests/functional/glusterd/test_detach_node_used_to_mount.py
nishith-vihar: Ack

tests/functional/glusterd/test_detach_node_used_to_mount.py
nishith-vihar: Ack

tests/functional/glusterd/test_detach_node_used_to_mount.py
nik-redhat: Nope we don't want to wait for it, we can simply stop it and proceed.

nishith-vihar: Ack

tests/functional/glusterd/test_detach_node_used_to_mount.py
nishith-vihar: Ack

nik-redhat:
If it's ready for review, then do mark it accordingly @nishith-vihar :)

nik-redhat (commented):
lgtm.

nik-redhat:
@srijan-sivakumar could you please test it in the CI?

aujjwal-redhat (approved):
LGTM

aujjwal-redhat (approved):

srijan-sivakumar:
@nik-redhat let me know if everything's fine. Blocked on your change request. I think those changes are taken care of. But still better to check it once.

nik-redhat (approved):
LGTM.

PR #641, opened 2021-06-08
Changing conditional bash from and

srijan-sivakumar:
Sometimes, the unmount will fail causing the mountpoint to not to be deleted. Hence changing the bash logic.


nishith-vihar (approved):
LGTM

PR #647, opened 2021-06-08
Fixed few TCs and updates ops to fail early

nik-redhat:
Fix:
The TC test_get_state_on_brick_unmount needed the volume to be `only created` and not started,
hence the unmount after the volume creation was logically correct and only the brick root should be unmounted
not the complete brick for a volume.

For the TC test_shared_storage, added terminate to disable `shared_storage` in case test fails midway,
else the /etc/fstab` file is not cleared in case of failure. Also, updated the shared_Storage ops to fail early in case
of a negative scenario.

### DIFF
--- a/common/ops/gluster_ops/shared_storage_ops.py
+++ b/common/ops/gluster_ops/shared_storage_ops.py
@@ -56,7 +56,9 @@ class SharedStorageOps(AbstractOps):
 
         return True
 
-    def is_shared_volume_mounted(self, node: str, timeout: int = 20) -> bool:
+    def is_shared_volume_mounted_or_unmounted(self, node: str,
+                                              is_mounted=True,
+                                              timeout: int = 20) -> bool:
         """
         Checks if shared storage volume is mounted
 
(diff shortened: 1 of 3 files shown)


nik-redhat:
Previously the TC test_shared_storage took around 300secs, and now it takes around 98 secs :p

Test run:
```
Table:
 test_shared_storage
 
|   Generic   |     PASS    | 98.97737884521484 |


Framework runtime : 109.23168206214905
```

```
Table:
 test_get_state_on_brick_unmount
 
|     rep     |     PASS    | 38.17700004577637  |
|     dist    |     PASS    | 47.14192605018616  |
|   dist-rep  |     PASS    | 49.77601075172424  |
|     disp    |     PASS    | 44.49259376525879  |
|  dist-disp  |     PASS    | 47.97344255447388  |
|     arb     |     PASS    | 43.70880627632141  |
|   dist-arb  |     PASS    | 48.588592767715454 |


Framework runtime : 329.4074890613556
```

srijan-sivakumar (approved):
LGTM

PR #648, opened 2021-06-08
Fix few TCs and ops

nik-redhat:
Fix:
Fixed small list indices error and return value check  error in the TCs.

### DIFF
--- a/tests/functional/glusterd/test_volume_create.py
+++ b/tests/functional/glusterd/test_volume_create.py
@@ -87,7 +87,7 @@ class TestCase(DParentTest):
         ret = redant.setup_volume(self.volume_name1, self.server_list[0],
                                   conf_dict, self.server_list,
                                   self.brick_roots, excep=False)
-        if ret['error_code'] == 0:
+        if not ret:
             raise Exception("Unexpected: Successfully created the volume with "
                             "already existing volname")
 
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -274,7 +274,7 @@ class BrickOps(AbstractOps):
 
         return ret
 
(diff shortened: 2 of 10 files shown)


nik-redhat:
Test run:
```
Table:
 test_peer_status
 
|   Generic   |     PASS    | 30.459044933319092 |


Framework runtime : 40.60669922828674
```
```
Table:
 test_shared_storage
 
|   Generic   |     PASS    | 100.54803228378296 |


Framework runtime : 112.68552374839783
```

This TC will fail in CI, because we are creating volume using `force` there.
```
Table:
 test_peer_probe
 
|   Generic   |     PASS    | 61.47437238693237 |


Framework runtime : 73.81182813644409
```

The rest of the TCs require 4 servers at least, hence not added the result.

nishith-vihar (approved):
LGTM

tests/functional/glusterd/test_volume_create.py
```diff
@@ -87,7 +87,7 @@ def run_test(self, redant):
         ret = redant.setup_volume(self.volume_name1, self.server_list[0],
                                   conf_dict, self.server_list,
                                   self.brick_roots, excep=False)
-        if ret['error_code'] == 0:
+        if not ret:
```
aujjwal-redhat, line 90: Just had a doubt. ret is the dictionary right?

nik-redhat: This is a corner case, and we are expecting the `gluster vol list` output to show the volume list contains the volume already, and that will return a bool. Hence, updated the check accordingly. If this TC fails at this point due to type error, then anyway there is something going wrong in some other part, so we are good here.

aujjwal-redhat: Oh ok. :+1:

Issue #667, opened 2021-06-16
Migrate test remove brick scenarios


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_remove_brick_scenarios.py)

PR #668, opened 2021-06-16
Migrated test_remove_brick_scenarios

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_remove_brick_scenarios.py)

Fixes: #667

Updates: #292

### DIFF
--- a/tests/functional/glusterd/test_remove_brick_scenarios.py
+++ b/tests/functional/glusterd/test_remove_brick_scenarios.py
@@ -20,22 +20,10 @@ Description:
 This test deals with scenarios with remove brick
 functionality.
 
-from multiprocessing.pool import ThreadPool
-from glusto.core import Glusto as g
-from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
-from glustolibs.gluster.exceptions import ExecutionError
-from glustolibs.gluster.brick_ops import remove_brick
-from glustolibs.gluster.brick_libs import get_all_bricks
-from glustolibs.gluster.rebalance_ops import (rebalance_start,
-                                              rebalance_status,
-                                              wait_for_fix_layout_to_complete)
-from glustolibs.gluster.glusterdir import mkdir
-from glustolibs.gluster.glusterfile import get_fattr
-
-@runs_on([['distributed-replicated'], ['glusterfs']])
 """
 # disruptive;dist-rep
 
+from multiprocessing.pool import ThreadPool
 from tests.d_parent_test import DParentTest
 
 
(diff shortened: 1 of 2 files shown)


aujjwal-redhat:
```js
✔ Environment setup successful.
✔ Running test case : test_remove_brick_scenarios-dist-rep
✔ test_remove_brick_scenarios-dist-rep Succeeded
Summary:
Table:
 test_remove_brick_scenarios
 
|   dist-rep  |     PASS    | 2602.0067014694214 |



Framework runtime : 2613.1035239696503

```

tests/functional/glusterd/test_remove_brick_scenarios.py
```diff
@@ -0,0 +1,150 @@
...
+
+        # Checking status of brick remove operation for other pair of bricks.
+        ret = redant.remove_brick(self.server_list[0], self.vol_name,
+                                  remove_brick_list_other_pair, 'status',
+                                  excep=False)
+        if ret['msg']['opRet'] == 0:
+            raise Exception("Error: Got status on other pair of bricks.")
+
+        # Stopping remove operation for non-existent bricks.
+        remove_brick_list_non_existent = [bricks_list[0] + 'non-existent',
```
srijan-sivakumar, line 103: use f strings for string concatenation.

srijan-sivakumar (changes requested):

tests/functional/glusterd/test_remove_brick_scenarios.py
aujjwal-redhat: Ack

aujjwal-redhat:
```js
Table:
 test_remove_brick_scenarios
 
|   dist-rep  |     PASS    | 2491.1046924591064 |



Framework runtime : 2501.6858842372894
```

srijan-sivakumar:
2k seconds!!

aujjwal-redhat:
> 2k seconds!!

yeah, the IO operation is quite heavy. It creates 100000 files and I guess gluster takes time in such large number of files.
For less number of files it took less time. Checked with 10, 100, 1000 and 10000 as well.

nik-redhat:
If we use a lesser number of files will that be a problem? If not let's reduce the number to 10000 and see if that works.

aujjwal-redhat:
> If we use a lesser number of files will that be a problem? If not let's reduce the number to 10000 and see if that works.

yeah 10000 works as well and it takes less time. But will that be fine for the test case logic coz 100000 files are also working. Just the time is too much. :-|

srijan-sivakumar:
> > If we use a lesser number of files will that be a problem? If not let's reduce the number to 10000 and see if that works.
> 
> yeah 10000 works as well and it takes less time. But will that be fine for the test case logic coz 100000 files are also working. Just the time is too much. :-|

Technically the idea of creating large number of files is to have file entries under all bricks and not just some bricks. This is because of how gluster works.
1. With the help of hash determined from the file, it is decided which subvolume that file will reside, though in case of the dirs, they are placed on all subvolumes.
2. Coming to layout, which we can simply understand as the range of hashes attached to each brick and files whose hash falls under that range will be stored in those said bricks. Add and remove brick in this way would re-trigger how the layout is assigned to all bricks, hence the need to re-trigger fix-layout which will redo the calculations and assign new layouts, but that'd mean some files will exist in bricks where they're now not supposed to. Now what I'm still unsure of is whether there is a separate migrate phase or would the fix-layout itself take care of the migrating of files. @nik-redhat any idea ?

nik-redhat:
So, we have two ways to trigger rebalance for a volume:
1. `gluster v rebal VOLNAME fix-layout start`: This will only update the layout of the volumes in the process and won't migrate the data.
2. `gluster v rebal VOLNAME start`: This uses the `fix-layout and migration` strategy and does the above task as well as migrates the data based on the storage of the bricks, uniformly.

So, generally what we do is the second option, and the same is being done in our OPs. I think we can go ahead with the TC and run it for 100000 files. What say?

srijan-sivakumar:
> So, we have two ways to trigger rebalance for a volume:
> 
>     1. `gluster v rebal VOLNAME fix-layout start`: This will only update the layout of the volumes in the process and won't migrate the data.
> 
>     2. `gluster v rebal VOLNAME start`: This uses the `fix-layout and migration` strategy and does the above task as well as migrates the data based on the storage of the bricks, uniformly.
> 
> 
> So, generally what we do is the second option, and the same is being done in our OPs. I think we can go ahead with the TC and run it for 100000 files. What say?

As discussed offline ( off-github ), going ahead with this TC.

srijan-sivakumar (approved):

Issue #675, opened 2021-06-18
Add more vol ops


aujjwal-redhat:
* log_volume_info_and_status
* is_volume_exported

PR #677, opened 2021-06-18
Added more volume ops

aujjwal-redhat:
Added log_volume_info_nad_status and is_volume_exported.

Fixes: #675

### DIFF
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -1105,13 +1105,55 @@ class VolumeOps(AbstractOps):
             status is successful. False Otherwise.
         """
         ret = self.get_volume_info(node, volname, False)
-        if int(ret['msg']['opRet']) != 0:
+        if 'msg' in ret.keys() and int(ret['msg']['opRet']) != 0:
             return False
 
         ret = self.get_volume_status(volname, node,
                                      excep=False)
-        if int(ret['msg']['opRet']) != 0:
+        if 'msg' in ret.keys() and int(ret['msg']['opRet']) != 0:
             return False
 
         return True
 
+    def is_volume_exported(self, node: str, volname: str,
+                           share_type: str):
+        """
+        Checks whether the volume is exported as nfs
+        or cifs share
+
+        Args:
+            node (str): Node on which cmd has to be executed.
+            volname (str): volume name
+            share_type (str): nfs or cifs
+
+        Returns:
+            bool: If volume is exported returns True.
+                False Otherwise.
+        """
+        if 'nfs' in share_type:
+            cmd = "showmount -e localhost"
+            self.execute_abstract_op_node(cmd, node)
+
+            cmd = f"showmount -e localhost | grep -w {volname}"
+            ret = self.execute_abstract_op_node(cmd,
+                                                node,
+                                                False)
+            if ret['error_code'] != 0:
+                return False
+            else:
+                return True
+
+        if 'cifs' in share_type or 'smb' in share_type:
+            cmd = "smbclient -L localhost"
+            self.execute_abstract_op_node(cmd, node)
+
+            cmd = ("smbclient -L localhost -U | "
+                   f"grep -i -Fw gluster {volname}")
+            ret = self.execute_abstract_op_node(cmd,
+                                                node,
+                                                False)
+            if ret['error_code'] != 0:
+                return False
+            else:
+                return True
+        return True
--- a/common/ops/gluster_ops/volume_ops.py
+++ b/common/ops/gluster_ops/volume_ops.py
@@ -425,12 +425,17 @@ class VolumeOps(AbstractOps):
             self.volume_stop(volname, node, True)
             self.volume_delete(volname, node)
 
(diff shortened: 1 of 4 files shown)


common/ops/gluster_ops/volume_ops.py
```diff
@@ -485,7 +490,10 @@ def get_volume_info(self, node: str = None, volname: str = 'all') -> dict:
 
         cmd = f"gluster volume info {volname} --xml"
 
-        ret = self.execute_abstract_op_node(cmd, node)
+        ret = self.execute_abstract_op_node(cmd, node, excep)
+
+        if not excep and int(ret['msg']['opRet']) != 0:
```
nik-redhat, line 495: opRet will have a string value, make it '0'

nik-redhat: Actually, we don't need this an `excep` flag for this function. Just an empty dict check in the caller function shall do, if it doesn't work then something wrong with the command.

common/ops/gluster_ops/volume_ops.py
```diff
@@ -1082,3 +1090,70 @@ def bulk_volume_creation(self, node: str, number_of_volumes: int,
...
+            self.execute_abstract_op_node(cmd, node)
+
+            cmd = ("smbclient -L localhost -U | "
+                   f"grep -i -Fw gluster {volname}")
+            ret = self.execute_abstract_op_node(cmd,
+                                                node,
+                                                False)
+            if ret['error_code'] != 0:
+                return False
+            else:
```
nik-redhat, line 1157: Same as above

common/ops/gluster_ops/volume_ops.py
```diff
@@ -1082,3 +1090,70 @@ def bulk_volume_creation(self, node: str, number_of_volumes: int,
...
+            self.execute_abstract_op_node(cmd, node)
+
+            cmd = f"showmount -e localhost | grep -w {volname}"
+            ret = self.execute_abstract_op_node(cmd,
+                                                node,
+                                                False)
+            if ret['error_code'] != 0:
+                return False
+            else:
+                return True
```
nik-redhat, line 1144: Remove this part and simply keep a `return True` at the end.

nik-redhat (changes requested):

common/ops/gluster_ops/volume_ops.py
aujjwal-redhat: Ack

common/ops/gluster_ops/volume_ops.py
aujjwal-redhat: Ack

common/ops/gluster_ops/volume_ops.py
aujjwal-redhat: Ack

common/ops/gluster_ops/volume_ops.py
aujjwal-redhat: But what if the command itself failed to execute. In that case there will be no dict.

nik-redhat: Ah....true. Then, let's keep this check, just update it to compare string.

aujjwal-redhat: yeah

nik-redhat:
Can you update the code for the `get_vol_status` as well, to not do any logging and return in case of `excep=False` in the same PR?

nik-redhat:
And, with that a couple oof TC needs fixes like `test_gluster_detect_drop_of_outbound_traffic` and `test_volume_status_with_absent_bricks` do fix them as well.

aujjwal-redhat:
> Can you update the code for the `get_vol_status` as well, to not do any logging and return in case of `excep=False` in the same PR?

Ok

aujjwal-redhat:
```js
Summary:
Table:
 test_volume_status_with_absent_bricks
 
|     dist    |     PASS    | 24.71252703666687  |
|     rep     |     PASS    | 32.111796855926514 |
|   dist-rep  |     PASS    | 34.945164918899536 |
|     disp    |     PASS    | 31.519803524017334 |
|  dist-disp  |     PASS    | 36.63633751869202  |
|     arb     |     PASS    | 32.81458806991577  |
|   dist-arb  |     PASS    | 36.21625828742981  |
```

tests/functional/glusterd/test_gluster_detect_drop_of_outbound_traffic.py
```diff
@@ -81,7 +81,7 @@ def run_test(self, redant):
         # Check volume status command
         vol_status = redant.get_volume_status(self.vol_name, self.server1,
                                               excep=False)
-        if vol_status is None:
+        if ret['msg']['opRet'] != '0':
             raise Exception("Failed to get the vol status")
```
nik-redhat, line 85: As per the above comment, it should fail, isn't it?

aujjwal-redhat: yeah True. Ig the logic needs to be changed here. on it

nik-redhat (approved):

Issue #687, opened 2021-06-22
Migrate test_glusterd_memory_consumption


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_glusterd_memory_consumption_increase.py)

PR #688, opened 2021-06-22
Migrated test_glusterd_memory_consumption_increase

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_glusterd_memory_consumption_increase.py).

Updates: #292

Fixes: #687

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -58,4 +58,4 @@ excluded_tests:
     - tests/functional/glusterd/test_volume_set_when_glusterd_stopped_on_one_node.py
     - tests/functional/glusterd/test_gluster_volume_status_xml_dump.py
     - tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py
-    - tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
+    - tests/functional/glusterd/test_glusterd_memory_consumption_increase.py # 100 volumes created, run multiple times
--- a/config/config.yml
+++ b/config/config.yml
@@ -58,3 +58,4 @@ excluded_tests:
     - tests/functional/glusterd/test_volume_set_when_glusterd_stopped_on_one_node.py
     - tests/functional/glusterd/test_gluster_volume_status_xml_dump.py
     - tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py
+    - tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
--- a/tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
+++ b/tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
@@ -73,21 +73,28 @@ class TestCase(DParentTest):
 
     #     g.log.info("Successfully deleted all the volumes")
 
-    # def _memory_consumption_for_all_nodes(self, pid_list):
-    #     """Fetch the memory consumption by glusterd process for
-    #        all the nodes
-    #     """
-    #     memory_consumed_list = []
-    #     for i, server in enumerate(self.servers):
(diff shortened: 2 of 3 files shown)


tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
```diff
@@ -0,0 +1,152 @@
...
+        # Create and start 100 volumes in a loop
+        volname = f"{self.test_name}{index}"
+        conf_dict = self.vol_type_inf[self.conv_dict['dist-rep']]
+        ret = self.redant.bulk_volume_creation(self.server_list[0], 100,
+                                               volname, conf_dict,
+                                               self.server_list,
+                                               self.brick_roots)
+
+        if not ret:
+            print("Failed to create volumes")
```
nik-redhat, line 39: Ig it should be an exception instead :)

nik-redhat:
@aujjwal-redhat can you try to run the test by changing the count to 10 volumes and confirm if the test is fine and add the results here.

aujjwal-redhat:
> @aujjwal-redhat can you try to run the test by changing the count to 10 volumes and confirm if the test is fine and add the results here.

Yeah.

tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
aujjwal-redhat: Oh yeah :sweat_smile:

aujjwal-redhat:
```js
Table:
 test_glusterd_memory_consumption_increase
 
|   Generic   |     PASS    | 617.1890532970428 |



Framework runtime : 627.4527649879456

✔ Tearing down successful.
```

nik-redhat:
@srijan-sivakumar shall we add this TC to the excluded list? It creates, starts, stops 100 volumes multiple times.

srijan-sivakumar:
> @srijan-sivakumar shall we add this TC to the excluded list? It creates, starts, stops 100 volumes multiple times.

Yes, I think that would be good. But can we also add a comment in the yaml file stating the reason behind exclusion. That would help if someone new picks and takes a look at the yaml.

nik-redhat:
> > @srijan-sivakumar shall we add this TC to the excluded list? It creates, starts, stops 100 volumes multiple times.
> 
> Yes, I think that would be good. But can we also add a comment in the yaml file stating the reason behind exclusion. That would help if someone new picks and takes a look at the yaml.

Cool.... @aujjwal-redhat please do the needful in this PR.

aujjwal-redhat:
> > @srijan-sivakumar shall we add this TC to the excluded list? It creates, starts, stops 100 volumes multiple times.
> 
> Yes, I think that would be good. But can we also add a comment in the yaml file stating the reason behind exclusion. That would help if someone new picks and takes a look at the yaml.

Ok adding it in the exclude tests

aujjwal-redhat:
@nik-redhat @srijan-sivakumar added it in excluded tests

config/config.yml
```diff
@@ -58,3 +58,4 @@ excluded_tests:
     - tests/functional/glusterd/test_volume_set_when_glusterd_stopped_on_one_node.py
     - tests/functional/glusterd/test_gluster_volume_status_xml_dump.py
     - tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py
+    - tests/functional/glusterd/test_glusterd_memory_consumption_increase.py
```
nik-redhat, line 61: Do add a comment as well, specifying the reason.

aujjwal-redhat: Aah... sorry, adding now.

nik-redhat (approved):

PR #703, opened 2021-06-24
UI improvements / refinements

csabahenk:
- Perform ssh as current (possibly unprivileged) user
- Factor out ssh connection routine into dedicated function
- Omit loading private keys for ssh connection
- Tolerate missing 'excluded_tests' key in config file
- Sanitize conf file loading
- Sanitize handling of fatal conditions
- Symlink log dir of latest run to 'latest'
- Make message about invalid log level more helpful

### All Submissions:

Desciption:

(If it closes the issue)
Fixes: #
(If it just updates the Issue)
Updates: #

### DIFF
--- a/core/redant_main.py
+++ b/core/redant_main.py
@@ -49,6 +49,9 @@ def pars_args():
     parser.add_argument("-xls", "--excel-sheet",
                         help="Spreadsheet for result. Default value is NULL",
                         dest="excel_sheet", default=None, type=str)
+    parser.add_argument("--show-backtrace",
+                        help="Show full backtrace on error",
+                        dest="show_backtrace", action='store_true')
     return parser.parse_args()
 
 
@@ -64,14 +67,23 @@ def main():
     start = time.time()
     args = pars_args()
 
+    if args.show_backtrace:
+        def errer(exc, msg=None):
+            raise exc
+    else:
+        def errer(exc, msg=None):
+            if not msg:
+                msg = "error: {exc}"
+            print(msg.format(exc=exc), file=sys.stderr)
+            sys.exit(1)
+
     spinner = Halo(spinner='dots')
     spinner.start("Starting param handling")
     try:
         param_obj = ParamsHandler(args.config_file)
     except OSError as e:
-        print("Error on loading config file:", e)
         spinner.fail("error in param handling")
-        sys.exit(1)
+        errer(e, "Error on loading config file: {exc}")
     spinner.succeed("Param Handling Success.")
 
     spinner.start("Building test list")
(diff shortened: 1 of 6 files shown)


nik-redhat (approved):
LGTM

aujjwal-redhat (approved):
LGTM

srijan-sivakumar:
I've triggered the CI run with this patch as we still don't have a GH linked CI check..will update it post the run.

srijan-sivakumar:
```
Summary:
+---------------+-------+-------------------+
|    Category   | Cases |    Pass Percent   |
+---------------+-------+-------------------+
| nonDisruptive |   22  | 86.15384615384616 |
|   Disruptive  |   63  | 76.10062893081762 |
|     Total     |   85  | 79.01785714285714 |
+---------------+-------+-------------------+
```

Found no issues in the CI runs. The failures were expected issues because of Test cases.

core/redant_main.py
```diff
@@ -82,29 +95,34 @@ def main():
...
 
     spinner.start("Creating log dirs")
     # Creating log dirs.
-    args.log_dir = f'{args.log_dir}/{datetime.datetime.now()}'
-    Logger.log_dir_creation(args.log_dir, TestListBuilder.get_test_path_list())
+    current_time_rep = str(datetime.datetime.now())
+    log_dir_current = f"{args.log_dir}/{current_time_rep}"
+    Logger.log_dir_creation(
+        log_dir_current, TestListBuilder.get_test_path_list())
+    latest = 'latest'
```
srijan-sivakumar, line 109: nit, I feel we can directly use `latest` as a string in both instances, i.e. inside line 110 as well as 112. That'll reduce the run time evaluation to be done and also the requirement for one more variable. Though it won't give any performance boost as such, hence nitpicking.

csabahenk: I thought that the Python compiler is good enough at optimization that using repeated literals or a local variable does not matter, the resultant bytecode will be the same. This might be the case, or not... anyway, it's not a hot code path, so it's not that part of the code where there is much use of discussing optimization.

It's more like a question of coding principles, style or personal preferences.

The principle I followed: try to avoid inputting literal values repeatedly.

srijan-sivakumar: > I thought that the Python compiler is good enough at optimization that using repeated literals or a local variable does not matter, the resultant bytecode will be the same. This might be the case, or not... anyway, it's not a hot code path, so it's not that part of the code where there is much use of discussing optimization.

Ack. Even I'm not sure about the level of optimization the compiler would be undertaking. But as you mentioned, it is not a critical section code and also even if the compiler wouldn't have optimized it, the said improvement wouldn't have been a great improvement. 

> It's more like a question of coding principles, style or personal preferences.
> 
> The principle I followed: try to avoid inputting literal values repeatedly.

Gotcha :)

Issue #728, opened 2021-06-29
Add heal ops required in arbiter


aujjwal-redhat:
Add the following heal ops:
* enable_self_heal_daemon
* disable_self_heal_daemon
* do_bricks_exist_in_shd_volfile

PR #729, opened 2021-06-29
Added heal ops required for arbiter

aujjwal-redhat:
Added the following heal ops:
* enable_self_heal_daemon
* disable_self_heal_daemon
* do_bricks_exist_in_shd_volfile

Fixes: #728

### DIFF
--- a/common/ops/gluster_ops/heal_ops.py
+++ b/common/ops/gluster_ops/heal_ops.py
@@ -567,10 +567,10 @@ class HealOps:
                 if "option remote-host" in each_line:
                     host = each_line.split(" ")[2]
 
-        self.logger.info(f"Brick List from volume "
-                         f"info: {brick_list}")
-        self.logger.info("Brick List from glustershd server volume "
-                         f"file: {brick_list_server_vol}")
+        self.logger.debug(f"Brick List from volume "
+                          f"info: {brick_list}")
+        self.logger.debug("Brick List from glustershd server volume "
+                          f"file: {brick_list_server_vol}")
 
         if set(brick_list) != set(brick_list_server_vol):
             return False
--- a/common/ops/gluster_ops/heal_ops.py
+++ b/common/ops/gluster_ops/heal_ops.py
@@ -517,8 +517,7 @@ class HealOps:
 
     def do_bricks_exist_in_shd_volfile(self, volname: str,
                                        brick_list: list,
-                                       node: str,
-                                       path: str = None) -> bool:
+                                       node: str) -> bool:
         """
         Checks whether the given brick list is present in glustershd
         server volume file
@@ -528,14 +527,13 @@ class HealOps:
             brick_list (list): brick list of a volume which needs to
                                compare in glustershd server volume file
             node (str): Node on which commands will be executed.
-            path (str): Dynamic path to check the shd file. Default
-                        is None.
 
         Returns:
             bool : True if brick exists in glustershd server volume file.
                    False Otherwise
         """
         GLUSTERSHD = "/var/lib/glusterd/glustershd/glustershd-server.vol"
+        path = f"/var/lib/glusterd/vols/{volname}/{volname}-shd.vol"
         if not self.path_exists(node, GLUSTERSHD):
             if not self.path_exists(node, path):
                 self.logger.error("Can't find shd file")
--- a/common/ops/gluster_ops/heal_ops.py
+++ b/common/ops/gluster_ops/heal_ops.py
@@ -517,7 +517,8 @@ class HealOps:
 
     def do_bricks_exist_in_shd_volfile(self, volname: str,
                                        brick_list: list,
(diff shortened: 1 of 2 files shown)


aujjwal-redhat:
```sh
 test_heal_ops
 
|     rep     |     PASS    |        00:02:31       |     N/A     |

```

common/ops/gluster_ops/heal_ops.py
```diff
@@ -474,3 +474,106 @@ def is_shd_daemon_running(self, node: str,
...
+        ret = self.execute_abstract_op_node(cmd, node, False)
+        if ret['error_code'] != 0:
+            return False
+
+        return True
+
+    def do_bricks_exist_in_shd_volfile(self, volname: str,
+                                       brick_list: list,
+                                       node: str,
+                                       path: str = None) -> bool:
```
srijan-sivakumar, line 521: The path shouldn't be asked from the user. With the help of the volume name, we can create the path here itself.

common/ops/gluster_ops/heal_ops.py
```diff
@@ -474,3 +474,106 @@ def is_shd_daemon_running(self, node: str,
...
+                if parse:
+                    brick_list_server_vol.append(f"{host}:{brick}")
+                parse = False
+            elif parse:
+                if "option remote-subvolume" in each_line:
+                    brick = each_line.split(" ")[2]
+                if "option remote-host" in each_line:
+                    host = each_line.split(" ")[2]
+
+        self.logger.info(f"Brick List from volume "
```
srijan-sivakumar, line 572: I think we can keep this log as debug. Same for the one following this.

tests/example/sample_component/test_heal_ops.py
```diff
@@ -34,26 +34,47 @@ def _perform_simple_io(self):
...
+                                        self.server_list[0])):
+            raise Exception("Failed to enable the self-heal daemon")
+
         self._perform_simple_io()
         if self.heal_info is None:
             raise Exception("Failed to get heal info")
 
+        bricks_list = redant.get_all_bricks(self.vol_name,
+                                            self.server_list[0])
+        path = (f"/var/lib/glusterd/vols/{self.vol_name}/"
```
srijan-sivakumar, line 70: As mentioned, path of volfile should be generated dynamically using the volname in the ops function rather than being passed from the test case by the user.

srijan-sivakumar (dismissed):

common/ops/gluster_ops/heal_ops.py
aujjwal-redhat: Ack

tests/example/sample_component/test_heal_ops.py
aujjwal-redhat: Oh I thought by dynamic path you meant giving user the control over the path. Will change that.

aujjwal-redhat:
```sh

 test_heal_ops
 
|     rep     |     PASS    |        00:02:29       |     N/A     |
```

common/ops/gluster_ops/heal_ops.py
```diff
@@ -474,3 +474,104 @@ def is_shd_daemon_running(self, node: str,
...
+            volname : Name of the volume
+            node : Node on which commands are executed
+
+        Returns:
+            bool : True if setting self_heal_daemon option to
+                   'on' is successful. False otherwise.
+        """
+        cmd = f"gluster volume set {volname} self-heal-daemon on"
+        ret = self.execute_abstract_op_node(cmd, node, False)
+        if ret['error_code'] != 0:
```
nik-redhat, line 493: I doubt if `error_code` will contain the correct value here because there are a lot of places where the error_code is not updated by Gluster. So, better to execute the above command with `--xml` flag and check for `ret['msg']['opRet']` over here.

common/ops/gluster_ops/heal_ops.py
```diff
@@ -474,3 +474,104 @@ def is_shd_daemon_running(self, node: str,
...
+        Args:
+            volname : Name of the volume
+            node : Node on which commands are executed
+
+        Returns:
+            bool : True if setting self_heal_daemon option to
+                   'off' is successful. False otherwise.
+        """
+        cmd = f"gluster volume set {volname} self-heal-daemon off"
+        ret = self.execute_abstract_op_node(cmd, node, False)
```
nik-redhat, line 512: Same as above

nik-redhat (changes requested):

common/ops/gluster_ops/heal_ops.py
aujjwal-redhat: Ack

common/ops/gluster_ops/heal_ops.py
aujjwal-redhat: Ack

nik-redhat (approved):

Issue #734, opened 2021-06-30
Migrate test_cyclic_brick_kill_list.py


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py)

PR #738, opened 2021-06-30
Migrated test_cyclic_brick_kill_list

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py)

Updates: #292

Fixes: #734

### DIFF
--- a/tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py
+++ b/tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py
@@ -176,11 +176,11 @@ class ListMount(GlusterBaseClass):
             g.log.info('All self-heal Daemons are online')
 
         # Validate IO
-        g.log.info("Wait for IO to complete and validate IO ...")
-        ret = validate_io_procs(self.all_mounts_procs, self.mounts)
-        self.assertTrue(ret, "IO failed on some of the clients")
+        self.assertTrue(
+            validate_io_procs(self.all_mounts_procs, self.mounts),
+            "IO failed on some of the clients"
+        )
         self.io_validation_complete = True
-        g.log.info("IO is successful on all mounts")
 
         # Checking volume status
         g.log.info("Logging volume info and Status after bringing bricks "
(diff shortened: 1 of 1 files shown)


aujjwal-redhat:
```sh
Table:
 test_cyclic_brick_kill_list
 
|     rep     |     PASS    |        00:05:42       |     N/A     |
|   dist-rep  |     PASS    |        00:11:20       |     N/A     |

Framework runtime : 00 hours 17 minutes 12 seconds
```

tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py
```diff
@@ -0,0 +1,128 @@
...
+            bs_fil = f"test_brick_down_from_client_{mount_obj['client']}.txt"
+            proc = redant.create_files(num_files=2,
+                                       fix_fil_size="1M",
+                                       path=mount_obj['mountpath'],
+                                       node=mount_obj['client'],
+                                       base_file_name=bs_fil)
+            self.list_of_procs.append(proc)
+
+        # Validate IO
+        ret = redant.validate_io_procs(self.list_of_procs, self.mnt_list)
```
srijan-sivakumar, line 69: Internally validate_io_procs is checking if the async process has completed or not and waits till it is completed. But this isn't desired here. We just need to verify that the IO hasn't ended before starting with the kill brick and restart brick operation. But with the current logic in `validate_io_procs` the validation would mean either the IO is completed or IO failed. There's no status for IO is going on.. We do have a function in rexe called check_async_command_status which checks if the async operation has been completed or not. This would be a better candidate here than `validate_io_procs`.

srijan-sivakumar (changes requested):

tests/functional/arbiter/brick_cases/test_cyclic_brick_kill_list.py
aujjwal-redhat: Ack

aujjwal-redhat:
```sh
Table:
 test_cyclic_brick_kill_list
 
|     rep     |     PASS    |        00:05:42       |     N/A     |
|   dist-rep  |     PASS    |        00:11:12       |     N/A     |

Framework runtime : 00 hours 17 minutes 02 seconds

✔ Tearing down successful.
```

srijan-sivakumar (approved):

PR #747, opened 2021-07-01
Changing default config value for dist volume

srijan-sivakumar:
Changing the distributed count for a distributed volume in config file from 4->3

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -22,7 +22,7 @@ clients_info:
 #               each volume type
 volume_types:
     dist:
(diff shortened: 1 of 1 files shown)


nik-redhat:
Any specific reason to do so? 🤔

srijan-sivakumar:
> Any specific reason to do so? thinking

Two prong...
1. dist-3 is the most common distribute volume configuration
2. resource usage can be reduced.

nik-redhat (approved):

Issue #719, opened 2021-06-28
Restructuring and cleanup of Result handler


srijan-sivakumar:
There is a good amount of duplicate logic across the result handler. This needs to be cleaned up.

srijan-sivakumar:
Basically, what needs to be improved is the following,
1. Removing the duplicate computations performed.
2. Spreadsheet layout has to be changed from a format mimicing that of the console output to a proper format which will help people to easily analyse the data. This would mean inclusion of more information such as the component so as to divide test data component wise in the workbooks.
3. No overlapping of features. Currently even when the user requests for xls format, we get a console summary. This shouldn't be the case.

srijan-sivakumar:
Time format correction and the skip count corrected.

srijan-sivakumar:
```json
{
	"component":    {
						"tcNature1" : {
										"testName1" : { "volType1" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 2321.123 },
										                "volType2" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 3423.234 }
										              },
										"testName2" : { "volType1" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 2321.123 },
										                "volType2" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 3423.234 }
										              }
									 },
					    "tcNature2" : {
										"testName1" : { "volType1" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 2321.123 },
										                "volType2" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 3423.234 }
										              },
										"testName2" : { "volType1" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 2321.123 },
										                "volType2" : { "testResult" : "PASS/FAIL/SKIP", "timeTaken": 3423.234 }
										              }
									 }
					}
}
```

srijan-sivakumar:
The test result queue data is to be transformed into this format so as to have a better control when aggregating data as well as when creating the result for output, be it stdout or xls.

PR #750, opened 2021-07-02
[Result] Fixing skip count and time output

srijan-sivakumar:
The skip count took into account the total
subtest skips and hence the data outputted
was more than what was expected. This fix
changes how the count is aggregated.

Secondly, the time output was messing up whenever
the minute or hour wasn't an integer. Hence setting
them to int.

Updates: #719


aujjwal-redhat (approved):
LGTM

Issue #756, opened 2021-07-05
Migrate test_verify_df_output.py


aujjwal-redhat:
Migrate the [test case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_verify_df_output.py)

PR #757, opened 2021-07-05
Migrated test_verify_df_output

aujjwal-redhat:
Migrated the [test
case](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/glusterd/test_verify_df_output.py)

Updates: #292

Fixes: #756

### DIFF
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -362,7 +362,7 @@ class BrickOps(AbstractOps):
                 brick_iter += 1
 
             if len(unused_servers) > 0:
-                server_val = unused_servers[iteration]
+                server_val = unused_servers[0]
                 unused_servers.remove(server_val)
             else:
                 server_val = server_list[server_iter]
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -167,6 +167,11 @@ class BrickOps(AbstractOps):
 
         ret = self.execute_abstract_op_node(cmd, node, excep)
 
+        if not excep and ret['msg']['opRet'] != '0':
+            return ret
+
+        self.es.replace_brick_from_brickdata(volname, src_brick,
+                                             dest_brick)
         return ret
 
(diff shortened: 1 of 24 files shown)


aujjwal-redhat:
@srijan-sivakumar @nik-redhat I have added the logic. Lemme know if something has to be changed.

common/ops/gluster_ops/brick_ops.py
```diff
@@ -203,17 +208,17 @@ def replace_brick_from_volume(self, volname: str, node: str,
...
         subvols_list = self.get_subvols(volname, node)
 
+        if not src_brick:
+            # Randomly select a brick to replace
+            src_brick = (random.choice(random.choice(subvols_list)))
         if not dst_brick:
             _, dst_brick = self.form_brick_cmd(servers, brick_roots,
                                                volname, 1, True)
 
+            dst_brick = f"{src_brick.split(':')[0]}:{dst_brick.split(':')[1]}"
```
nik-redhat, line 218: Why this change?
This means it becomes necessary to use the same node for the new brick, but why?

aujjwal-redhat: The issue is before replace brick from volume the bricks were :

servera: [/path/to/brick/brick-0], serverb: [/path/to/brick/brick-1], serverc: [/path/to/brick/brick-1]

After replace it came out as :
servera: [/path/to/brick/brick-3, /path/to/brick/brick-4], serverb: [/path/to/brick/brick-5], serverc: []

And in this case df -h gave 4GB instead of 8GB normally...

I tested the same manually as shared in the redant grp

But when it is properly distributed it gave 8GB ... that's why I am considering that if in case dst brick is not given it's better to replace with a brick on the same server...

srijan-sivakumar: I think this might not work with a real scenario. For instance, if we had had a disk for each brick, we would not face this issue. Hence I feel we shouldn't restrict the replace to the same node.

aujjwal-redhat: Yeah .. true

nik-redhat:
Closing this wrt #929

PR #785, opened 2021-07-08
Used f strings in rmdir

aujjwal-redhat:
Used f-strings in place of + in rmdir function.

### DIFF
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -1125,9 +1125,9 @@ class IoOps(AbstractOps):
         cmd = "rmdir"
         if force:
             cmd = "rm"
-            cmd += " -rf"
+            cmd = f"{cmd} -rf"
 
-        cmd += f" {fqpath}"
+        cmd = f"{cmd} {fqpath}"
         ret = self.execute_abstract_op_node(cmd, node, False)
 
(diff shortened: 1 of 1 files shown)


common/ops/support_ops/io_ops.py
```diff
@@ -1125,9 +1125,9 @@ def rmdir(self, fqpath: str, node: str, force: bool = False) -> bool:
         cmd = "rmdir"
         if force:
             cmd = "rm"
-            cmd += " -rf"
+            cmd = f"{cmd} -rf"
```
nik-redhat, line 1128: This change is not needed actually, using + logically does the same thing.

nik-redhat (approved):

common/ops/support_ops/io_ops.py
aujjwal-redhat: just to maintain the standard : )

srijan-sivakumar (approved):

common/ops/support_ops/io_ops.py
srijan-sivakumar: Well technically f strings give performance improvement only when there are more than one string being concatenated. In this case..it'll be not at all an improvement. But as mentioned by you...this is a coding standard used by us...so let's continue doing so...

PR #793, opened 2021-07-09
Migrated test_afr_cli_no_splitbrain_resolution.py

aujjwal-redhat:
Migrated
[test_afr_cli_no_splitbrain_resolution.py](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py)

Updates: #292

### DIFF
--- a/tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py
+++ b/tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py
@@ -1,233 +1,197 @@
-#  Copyright (C) 2017-2021  Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License along
-#  with this program; if not, write to the Free Software Foundation, Inc.,
-#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+Copyright (C) 2017-2021  Red Hat, Inc. <http://www.redhat.com>
 
-# pylint: disable=too-many-statements, too-many-locals, unused-variable
-from glusto.core import Glusto as g
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


nik-redhat:
Output?

aujjwal-redhat:
```sh
Table:
 test_afr_cli_no_splitbrain_resolution
 
|     rep     |     PASS    |        00:02:44       |     N/A     |

Framework runtime : 00 hours 02 minutes 55 seconds
```

aujjwal-redhat:
> Test run output?

Added : )

tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py
```diff
@@ -0,0 +1,184 @@
...
+            self.redant.execute_abstract_op_node(f"mkdir -p "
+                                                 f"{self.mountpoint}",
+                                                 client)
+            self.redant.volume_mount(self.server_list[0],
+                                     self.vol_name,
+                                     self.mountpoint, client)
+
+    def terminate(self):
+        try:
+            ret = self.redant.wait_for_io_to_complete(self.proc,
```
nik-redhat, line 56: Better to add a check for `self.proc` before using it.

aujjwal-redhat: But even if self.proc is empty it will work. That's why I have initialized self.proc in the beginning to prevent any failure.

nik-redhat: I know, but why even go inside the function there are no processes to check, hence better to have the check.

aujjwal-redhat: Ack. But there are already many tests where we did the same. Np, will make the change : )

aujjwal-redhat:
```sh
Table:
 test_afr_cli_no_splitbrain_resolution
 
|     rep     |     PASS    |        00:02:44       |     N/A     |

Framework runtime : 00 hours 02 minutes 54 seconds
```

aujjwal-redhat:
```sh
Table:
 test_afr_cli_no_splitbrain_resolution
 
|     rep     |     PASS    |        00:02:44       |     N/A     |

Framework runtime : 00 hours 02 minutes 54 seconds
```

tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py
nik-redhat: Oh ok. It is not mandatory, we can live with this code as well. It's just that I noticed now so thought we can have this check.

nik-redhat (approved):

tests/functional/afr/test_afr_cli_no_splitbrain_resolution.py
aujjwal-redhat: Yeah, a condition before is better actually. Not much but some computation reduced. : )

PR #817, opened 2021-07-14
[Common] Glusterd ungraceful kill added.

srijan-sivakumar:
Sometimes glusterd doesn't stop with systemctl
stop command, this causes the wait_till_glusterd_to_stop
to go on unable to do anything. Hence added an ungraceful
mechanism to handle this corner case in glusterd.

### DIFF
--- a/common/ops/gluster_ops/glusterd_ops.py
+++ b/common/ops/gluster_ops/glusterd_ops.py
@@ -233,7 +233,7 @@ class GlusterdOps(AbstractOps):
 
     def kill_glusterd_ungraceful(self, node: str):
         """
-        Method to kill glusterd, given its pid.
+        Method to kill glusterd, given its pid is active.
 
         Args:
             node (str|list): Node(s) on which this is to be executed.
--- a/common/ops/gluster_ops/glusterd_ops.py
+++ b/common/ops/gluster_ops/glusterd_ops.py
@@ -244,9 +244,7 @@ class GlusterdOps(AbstractOps):
 
         for nd in node:
             ret = self.execute_abstract_op_node(cmd1, nd, False)
-            if ret['error_code'] == 1:
-                continue
-            elif ret['error_code'] == 0:
+            if ret['error_code'] == 0:
                 pid = ret['msg'][0].strip()
                 cmd2 = f"kill -9 {pid}"
                 self.execute_abstract_op_node(cmd2, nd)
--- a/common/ops/gluster_ops/glusterd_ops.py
+++ b/common/ops/gluster_ops/glusterd_ops.py
@@ -231,6 +231,26 @@ class GlusterdOps(AbstractOps):
             count += 1
         return False
 
+    def kill_glusterd_ungraceful(self, node: str):
+        """
+        Method to kill glusterd, given its pid.
+
+        Args:
+            node (str|list): Node(s) on which this is to be executed.
+        """
+        if not isinstance(node, list):
+            node = [node]
+        cmd1 = "pidof glusterd"
+
+        for nd in node:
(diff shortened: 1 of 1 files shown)


common/ops/gluster_ops/glusterd_ops.py
```diff
@@ -231,6 +231,24 @@ def wait_for_glusterd_to_start(self, node: str, timeout: int = 80) -> bool:
             count += 1
         return False
 
+    def kill_glusterd_ungraceful(self, node: str):
+        """
+        Method to kill glusterd, given its pid.
```
nik-redhat, line 236: nit give its pid is still active (?)

srijan-sivakumar: I didn't get you ?

common/ops/gluster_ops/glusterd_ops.py
```diff
@@ -231,6 +231,24 @@ def wait_for_glusterd_to_start(self, node: str, timeout: int = 80) -> bool:
...
+            node (str|list): Node(s) on which this is to be executed.
+        """
+        if not isinstance(node, list):
+            node = [node]
+        cmd1 = "pidof glusterd"
+
+        for nd in node:
+            ret = self.execute_abstract_op_node(cmd1, nd, False)
+            if ret['error_code'] == 0:
+                pid = ret['msg'][0].strip()
```
nik-redhat, line 248: We need to strip for `\n` specifically, else it won't work.

common/ops/gluster_ops/glusterd_ops.py
```diff
@@ -231,6 +231,24 @@ def wait_for_glusterd_to_start(self, node: str, timeout: int = 80) -> bool:
...
+
+        Args:
+            node (str|list): Node(s) on which this is to be executed.
+        """
+        if not isinstance(node, list):
+            node = [node]
+        cmd1 = "pidof glusterd"
+
+        for nd in node:
+            ret = self.execute_abstract_op_node(cmd1, nd, False)
```
nik-redhat, line 246: Why are we using passing `False` here?

common/ops/gluster_ops/glusterd_ops.py
nik-redhat: I mean a better doc string would be:
`Method to kill glusterd, given its pid is till active`

common/ops/gluster_ops/glusterd_ops.py
srijan-sivakumar: strip will take care of spaces and `\n`.

common/ops/gluster_ops/glusterd_ops.py
srijan-sivakumar: ah..ok.

common/ops/gluster_ops/glusterd_ops.py
srijan-sivakumar: For one, I don't want exception handling to come into play here...

nik-redhat: If that's the case, we should pass `False` on line: 250 as well.

srijan-sivakumar: Technically the reason why I didn't want exception handling in running `pidof glusterd` is that if a node's gluster has properly shutdown, then the return value will be `1`. Now that is actually not an error but if exception handling is in force, that'll be treated as an error. Hence bypassing it in that code section. But, in the L250, I want the command to Pass and if the error code is other than 0, I want it to be treated as failure.

nik-redhat: Makes sense...

nik-redhat (approved):

PR #826, opened 2021-07-15
Migrated TC test_auth_allow_with_rebalance

nik-redhat:
Migrated TC [test_auth_allow_with_rebalance](https://github.com/gluster/glusto-tests/blob/master/tests/functional/authentication/test_auth_allow_with_rebalance.py)

Updates: #292 


nik-redhat:
This TC can be either kept on hold or can be let to fail, as there seems to be a bug in gluster with the `auth.allow` option when an add-brick operation is done.

aujjwal-redhat (approved):
The tc logic LGTM.

We can let it fail and add a note in the excel sheet as this is neither server issue nor logic issue.

srijan-sivakumar:
I think we can let it fail. A genuine bug should be shown as a failure here...so I think we should definitely merge this.

srijan-sivakumar (approved):

Issue #839, opened 2021-07-19
[Doc] Snapshot ops.


srijan-sivakumar:
Adding documentation for snapshot ops.

PR #846, opened 2021-07-19
[Doc] Adding Snap ops markdown

srijan-sivakumar:
Added snap ops markdown doc.

Fixes: #839


docs/BP/Ops/snap_ops.md
```diff
@@ -0,0 +1,513 @@
...
+                - error_msg: error message
+                - error_code: error code returned
+                - cmd : command that got executed
+                - node : node on which the command got executed
+                
+        Example:
+        ```python
+            redant.enable_uss(self.vol_name, self.server_list[0])
+		```
+
```
aujjwal-redhat, line 26: It is not able to display in correct format.

Preview:

Screenshot from 2021-07-21 10-21-50

aujjwal-redhat: Same is the case in the ones below as well.

aujjwal-redhat (changes requested):

docs/BP/Ops/snap_ops.md
srijan-sivakumar: I think it is not working in markdown. I just checked across the docs. The example is sticking out...This means we might need to also clear out from other docs.

srijan-sivakumar: @aujjwal-redhat I think we can take this in for now. We need to decide whether we are going to live with this type of documentation of shifting to readthedocs or github wiki in the coming days. Till then, I think this won't cause any confusion.

aujjwal-redhat: We can remove the *```python* completely then else anyone will get confused on seeing the examples part. Just a suggestion else we can keep it as it is.

nik-redhat: Hmm, then that can be made but not an urgent requirement I would say.

aujjwal-redhat: Yeah.

aujjwal-redhat (approved):

PR #852, opened 2021-07-21
Migrated test_existing_shd_should_take_care_of_self_healing

aujjwal-redhat:
Migrated
[test_existing_shd_should_take_care_of_self_healing](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py)

Updates: #292

### DIFF
--- a/tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
+++ b/tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
@@ -22,7 +22,6 @@ Description:
 # TODO: nfs, cifs
 
 import time
-import traceback
 from tests.d_parent_test import DParentTest
 
 
--- a/tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
+++ b/tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
@@ -29,7 +29,7 @@ from tests.d_parent_test import DParentTest
 
 class TestCase(DParentTest):
 
-    def test_existing_glustershd_should_take_care_of_self_healing(self):
+    def run_test(self, redant):
         """
         Test Script which verifies that the existing glustershd should take
         care of self healing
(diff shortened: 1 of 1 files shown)


aujjwal-redhat:
```sh
.
✔ Running test case : test_existing_shd_should_take_care_of_self_healing-rep
✔ test_existing_shd_should_take_care_of_self_healing-rep Succeeded
✔ Running test case : test_existing_shd_should_take_care_of_self_healing-dist-rep
✔ test_existing_shd_should_take_care_of_self_healing-dist-rep Succeeded
Results for heal
Test nature : disruptive
TestName : test_existing_shd_should_take_care_of_self_healing
|     rep     |     PASS    |        00:11:10       |      NA     |
|   dist-rep  |     PASS    |        00:09:48       |      NA     |
Total Time Taken : 00 hours 21 minutes 37 seconds
✔ Tearing down successful.
```

tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
```diff
@@ -0,0 +1,187 @@
...
+        try:
+            ret = self.redant.wait_for_io_to_complete(self.all_mounts_procs,
+                                                      self.mounts)
+            if not ret:
+                raise Exception("IO failed on some clients")
+        except Exception as error:
+            tb = traceback.format_exc()
+            self.redant.logger.error(error)
+            self.redant.logger.error(tb)
+        super().terminate()
```
nik-redhat, line 41: Since in this TC, we are validating the IO exactly after it is started, we don't need this check in the terminate...because if it failed there, it will fail here as well.

nik-redhat (changes requested):
Also, have you tried running the TC by removing the sleep or reducing the time duration?

aujjwal-redhat:
> Also, have you tried running the TC by removing the sleep or reducing the time duration?

No. It's taking a lot of time even without the sleep that's why I didn't try to remove that 90 s and 60s sleep.

tests/functional/afr/heal/test_existing_shd_should_take_care_of_self_healing.py
aujjwal-redhat: Ok

nik-redhat (approved):

PR #871, opened 2021-07-26
Migrated TC test_self_heal_50k_files_heal_command_by_add_brick

nik-redhat:
Migrated TC [test_self_heal_50k_files_heal_command_by_add_brick](https://github.com/gluster/glusto-tests/blob/master/tests/functional/arbiter/test_self_heal_50k_files_heal_command_by_add_brick.py)

Updates: #292 

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -60,3 +60,4 @@ excluded_tests:
     - tests/functional/glusterd/test_peer_probe_firewall_ports_not_opened.py
     - tests/functional/glusterd/test_glusterd_memory_consumption_increase.py # 100 volumes created, run multiple times
     - tests/functional/arbiter/test_self_heal_50k_files.py # Async process getting hanged while creating 50k files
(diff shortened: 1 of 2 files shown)


nik-redhat:
A similar TC that deals with 50k files.

srijan-sivakumar:
@nik-redhat I think it is better to open an issue on async timeout. So that we can plan and work on it.

nik-redhat:
> @nik-redhat I think it is better to open an issue on async timeout. So that we can plan and work on it.

I have opened issue #878 for the same.

nik-redhat:
I think we are good to merge this TC if the logic looks fine to you all. I will send out a separate PR for the timeout update.

srijan-sivakumar (approved):
@nik-redhat LGTM. We can test this properly once we have the timeout for async. For now, taking in this patch.

PR #888, opened 2021-07-27
Migrated TC test_brickcases

nik-redhat:
Migrated TC [test_brickcases](https://github.com/gluster/glusto-tests/blob/master/tests/functional/arbiter/brick_cases/test_brickcases.py)

Updates: #292 

### DIFF
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -327,6 +327,7 @@ class BrickOps(AbstractOps):
 
         iter_add = 0
         if add_flag:
+            brick = ""
             unused_servers = server_list[:]
             brick_list = self.get_all_bricks(volname, server_list[0])
             for brick in brick_list:
@@ -358,7 +359,7 @@ class BrickOps(AbstractOps):
                 server_iter = 0
                 brick_iter += 1
 
-            if len(unused_servers):
+            if len(unused_servers) > 0:
                 server_val = unused_servers[iteration]
                 unused_servers.remove(server_val)
             else:
--- a/common/ops/gluster_ops/brick_ops.py
+++ b/common/ops/gluster_ops/brick_ops.py
@@ -322,15 +322,33 @@ class BrickOps(AbstractOps):
         brick_cmd = ""
         server_iter = 0
         brick_iter = 0
+        unused_servers = []
+        last_node = ""
 
         iter_add = 0
         if add_flag:
+            unused_servers = server_list[:]
             brick_list = self.get_all_bricks(volname, server_list[0])
             for brick in brick_list:
                 brick_num = int(brick.split('-')[-1])
                 if iter_add < brick_num:
                     iter_add = brick_num
+
+                # Create unused servers list
+                for server in unused_servers:
+                    if server in brick:
+                        unused_servers.remove(server)
+
+            # Save the last used node, so that new brick addition starts from
+            # a different node
+            last_node = brick.split(':')[0]
+
             iter_add = iter_add + 1
+
+        # Don't begin brick creation from the same node as the last node
+        # If we are adding new bricks
+        if last_node == server_list[server_iter]:
+            server_iter += 1
 
         for iteration in range(mul_fac):
             # If no more servers left in brick_root dict, start from beginning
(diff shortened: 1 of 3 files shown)


nik-redhat:
```
Results for brick_cases
Test nature : disruptive
TestName : test_brickcases
|   Generic   |     PASS    |        00:02:25       |      NA     |
Total Time Taken : 00 hours 02 minutes 36 seconds
✔ Tearing down successful.
```

common/ops/gluster_ops/brick_ops.py
```diff
@@ -322,16 +322,35 @@ def form_brick_cmd(self, server_list: list, brick_root: dict,
...
             brick_list = self.get_all_bricks(volname, server_list[0])
             for brick in brick_list:
                 brick_num = int(brick.split('-')[-1])
                 if iter_add < brick_num:
                     iter_add = brick_num
+
+                # Create unused servers list
+                for server in unused_servers:
+                    if server in brick:
+                        unused_servers.remove(server)
```
srijan-sivakumar, line 341: This can lead to some undefined behaviour...or atleast strange behavior. Lists in python are tricky...I'd suggest you try running the code

```python
list_val = [1, 2, 3, 4]
for val in list_val:
    print(val, list_val)
    list_val.remove(val)
```
Try running the above code. You'd find the loop not running as we'd expect in C.

The best way here would be to use a separate variables for iteration in for loop and removal altogether, that too obtained by a deepcopy of unused_servers

srijan-sivakumar (changes requested):

common/ops/gluster_ops/brick_ops.py
srijan-sivakumar: For reference to how things work for python list during iteration -> [Link](https://blog.finxter.com/how-to-remove-items-from-a-list-while-iterating/)

nik-redhat: Interesting issue, will update the code.

nik-redhat:
```
Results for brick_cases
Test nature : disruptive
TestName : test_brickcases
|   Generic   |     PASS    |        00:02:36       |      NA     |
Total Time Taken : 00 hours 02 minutes 49 seconds
✔ Tearing down successful.
```

common/ops/gluster_ops/brick_ops.py
```diff
@@ -322,16 +322,37 @@ def form_brick_cmd(self, server_list: list, brick_root: dict,
...
 
+        if used_servers:
+            for server in server_list:
+                if server not in used_servers:
+                    unused_servers.append(server)
+
+        # Don't begin brick creation from the same node as the last node
+        # If we are adding new bricks
+        if last_node == server_list[server_iter]:
+            server_iter += 1
```
srijan-sivakumar, line 354: I'm not getting this logic..so basically, we are trying to prevent the new bricks from being added into the node that was the contained the last brick. And to prevent that we are increasing the server_iter by 1. But herein we are just comparing if that last node was the very first server in the server_list. What if say we have 3 servers and we have say 2, 2, 1. The last brick being in server 2. In this scenario, we'd be checking if last_node is the very first server in server_list. With it not being the last_node, we will add a new brick in server 1 itself. even though we should have say started with server 3 ( as it has lesser number of bricks ).

Just a thought...let me know if there is some other reasoning which I might be missing here.

srijan-sivakumar: Maybe the patch is about considering unused servers rather than optimizing with respect to brick weights ( i.e. number of bricks per server )

nik-redhat: > Maybe the patch is about considering unused servers rather than optimizing with respect to brick weights ( i.e. number of bricks per server )

Yes.....the scenario is that if we are creating bricks using fewer servers and have more servers left to consider while creating new bricks we should use that.
The above-mentioned scenario is not considered in this patch but can be done in a separate patch. Though, I don't have a logic yet on how to prevent all possible scenarios.

srijan-sivakumar: Cool. If possible, could you create a md inside RP for this change in brick ops. I feel we should start getting the design considerations into the RP directory so that if someone new wants to take things ahead, they can easily understand by glancing the doc once.

srijan-sivakumar: we can also take that as a separate issue for design updation and send separate set of patches to address that. Not blocking this PR for that updation..

srijan-sivakumar (approved):

Issue #698, opened 2021-06-24
Config file parsing sanitization


schaffung:
The config file parsing needs to be sanitized to determine whether the incoming values are enough to run tests of all volume types or not.

srijan-sivakumar:
This should also include whether the exclude path mentioned is valid or not. If there is an invalid path, the test suite execution needs to be stopped.

nik-redhat:
Seems like we have this check already.

PR #899, opened 2021-07-28
[Core] Exclude test path check

srijan-sivakumar:
Added check if the exclude test paths are valid.
If there is even a single invalid path, it is a cause
of worry and hence the framework will spit that out
as an error.

Updates: #698

### DIFF
--- a/core/redant_main.py
+++ b/core/redant_main.py
@@ -86,7 +86,12 @@ def main():
 
     spinner.start("Building test list")
     # Building the test list and obtaining the TC details.
-    excluded_tests = param_obj.get_excluded_tests()
+    excluded_result = param_obj.get_excluded_tests()
+    if not excluded_result[1]:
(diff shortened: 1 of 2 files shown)


aujjwal-redhat (approved):

core/redant_main.py
```diff
@@ -86,7 +86,12 @@ def errer(exc, msg=None):
 
     spinner.start("Building test list")
     # Building the test list and obtaining the TC details.
-    excluded_tests = param_obj.get_excluded_tests()
+    excluded_result = param_obj.get_excluded_tests()
+    if not excluded_result[1]:
+        spinner.fail("error in exclude list. Invalid path present")
```
nik-redhat, line 91: nit: Let's capitalize `E` in "error` as well.

nik-redhat (approved):

aujjwal-redhat (approved):

PR #904, opened 2021-07-29
minor changes

Sheetalpamecha:
This patch updates links, fix typos, correct commands

Fixes: #903

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,16 +80,16 @@ how to get things running and that's it, take the blue pill.
 
 ### STEP-BY-STEP procedure to run:
 1. git clone `[your fork for this repo]`
-2. Create a virtual environment : `virtualenv <virtual_env_name>`
+2. Create a virtual environment : `venv <virtual_env_name>`
 3. Activate the virtual-env : `source <virtual_env_name>/bin/activate`
 4. cd `[the-fork]`
 5. Run `pip3 install -r requirements.txt`
 6. To run the sample TC, just run the below cmd after populating the
 config file with relevant values. The command has to be run from the main redant
 reository. The tests path should be given with respect to the redant directory.
-`python3 ./core/redant_main.py -c ./core/parsing/config.yml -t tests/example/`
-7. Log files can be found at /tmp/redant.log [ default path ].
+`python3 ./core/redant_main.py -c ./config/config.yml -t tests/example/`
 For more options, run `python3 ./core/redant_main.py --help`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -80,16 +80,16 @@ how to get things running and that's it, take the blue pill.
 
 ### STEP-BY-STEP procedure to run:
 1. git clone `[your fork for this repo]`
-2. Create a virtual environment : `virtualenv <virtual_env_name>`
+2. Create a virtual environment : `venv <virtual_env_name>`
```
nik-redhat, line 83: I think this is version-dependent?

Sheetalpamecha: Yes, It is. But as python2 is deprecated, it is better to use python3 and its commands to reflect in the projects

nik-redhat: `virtualenv` works for both Python2 and Python3, so that won't cause any issue.
But, yes we can use python's in-built module `venv` as well. So, it will be better to keep both the commands.

Also, I think the command to create the virtual environment using `venv` is `python3 -m venv <virtual-env-name>`.

Sheetalpamecha: Yeah we can keep both

srijan-sivakumar: Yes, keeping both in the documentation would be better. The only difference between then is that venv is dependent more on python version while virtualenv is a lil independent.

Sheetalpamecha: For virtualenv an extra package need to be installed, while venv is a lib in python eliminating another installation If not used much.

Sheetalpamecha: > Also, I think the command to create the virtual environment using `venv` is `python3 -m venv <virtual-env-name>`.
Yes corrected the command and included both now

README.md
```diff
@@ -80,16 +80,18 @@ how to get things running and that's it, take the blue pill.
 
 ### STEP-BY-STEP procedure to run:
 1. git clone `[your fork for this repo]`
-2. Create a virtual environment : `virtualenv <virtual_env_name>`
+2. Create a virtual environment : 
+python2 `virtualenv <virtual_env_name>`
```
srijan-sivakumar, line 84: LGTM but I think we better not mention python2 in the doc. We can keep both the commands but not add any indication that redant works with python2. We can use the command `python3 -m virtualenv <virtual_env_name>`.

Sheetalpamecha: okay

nik-redhat (approved):

aujjwal-redhat (approved):

PR #915, opened 2021-08-02
Migrated TC test_arb_to_repl_conversion_with_io

nik-redhat:
Migrated TC [test_arb_to_repl_conversion_with_io](https://github.com/gluster/glusto-tests/blob/master/tests/functional/afr/test_arb_to_repl_conversion_with_io.py)

Updates: #292 


srijan-sivakumar:
@nik-redhat do we need to test this change or is there something else blocking this ?

nik-redhat:
Yeah, only the testing part is left. The heal is getting timed out for this TC.

nik-redhat:
The heal is not completing for the TC, even in the glusto env or for the glusto TC. Hence, marking it as excluded with proper justification.

srijan-sivakumar (approved):

Issue #911, opened 2021-08-02
Distro check in quorum_syslog.py


srijan-sivakumar:
The quorum_syslog test under glusterd component needs to have a check for distro as the syslog logging has been removed in latest fedora versions. Hence the test should not run if the distro is Fedora, it should be skipped.

PR #916, opened 2021-08-03
[Core][Common][Test] Exclude list fix

srijan-sivakumar:
1. Fixed the total runs count for a scenario when
all cases are skipped.
2. TEST_RES fix in a test.
3. Added check for distro in a case.
    
Fixes: #911

### DIFF
--- a/common/ops/support_ops/machine_ops.py
+++ b/common/ops/support_ops/machine_ops.py
@@ -178,14 +178,13 @@ class MachineOps(AbstractOps):
     def check_os(self, os_name: str, nodes: str,
                  os_version: str = None) -> bool:
         """
-        Checks the os release and compares the
-        os and version.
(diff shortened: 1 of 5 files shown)


common/ops/support_ops/machine_ops.py
```diff
@@ -202,10 +202,13 @@ def check_os(self, os_name: str, os_version: str, nodes: str):
                 return False
 
             out = item['msg']
-            if (
-                os_name not in out[0].lower()
-                or os_version not in out[1]
-            ):
+            if os_version is not None:
```
nik-redhat, line 205: nit: we can simply write: `if os_version:`

srijan-sivakumar: Ack..

nik-redhat (approved):

Issue #874, opened 2021-07-26
Clearing out old *.socket files from /var/run/gluster


srijan-sivakumar:
The socket files would keep on piling with the test cases. The best practice would be to clear a socket file once it has been used. But.. This would have to be done at two stages, before the framework begins and after the framework has completed the execution rather than during the suite run.

PR #920, opened 2021-08-03
[Common] Clearing out files and dirs

srijan-sivakumar:
Clearing out the old *.socket files and other dirs
from /var/run/gluster dir.

Fixes: #874

### DIFF
--- a/common/ops/support_ops/machine_ops.py
+++ b/common/ops/support_ops/machine_ops.py
@@ -146,10 +146,7 @@ class MachineOps(AbstractOps):
         cmd = ("rm -rf /var/lib/glusterd/vols/*; rm -rf /var/lib/glusterd"
                "/peers/*; rm -rf /var/lib/glusterd/snaps/*; "
                "rm -rf /var/lib/glusterd/glusterfind/.keys/*; "
-               "rm -rf /var/run/gluster/*.socket; "
-               "rm -rf /var/run/gluster/vols/*; "
-               "rm -rf /var/run/gluster/shd/*; "
-               "rm -rf /var/run/gluster/snaps/*")
+               "rm -rf /var/run/gluster/*; ")
         for node in server_list:
             self.execute_abstract_op_node(cmd, node, False)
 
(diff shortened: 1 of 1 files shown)


common/ops/support_ops/machine_ops.py
```diff
@@ -145,7 +145,11 @@ def hard_terminate(self, server_list: list, client_list: list,
         # directory on the servers.
         cmd = ("rm -rf /var/lib/glusterd/vols/*; rm -rf /var/lib/glusterd"
                "/peers/*; rm -rf /var/lib/glusterd/snaps/*; "
-               "rm -rf /var/lib/glusterd/glusterfind/.keys/*")
+               "rm -rf /var/lib/glusterd/glusterfind/.keys/*; "
+               "rm -rf /var/run/gluster/*.socket; "
+               "rm -rf /var/run/gluster/vols/*; "
+               "rm -rf /var/run/gluster/shd/*; "
+               "rm -rf /var/run/gluster/snaps/*")
```
nik-redhat, line 152: I think simply deleting everything using `/var/run/gluster*` should do instead of deleting it separately?

srijan-sivakumar: Not sure of the complete deletion. Because hard-terminate might be invoked if cleanup_volume fails in non disruptive tests. Also, the aim of this cleanup is to bring the file systems back to the state which is present when gluster is freshly installed.

Hence not deleting this dirs which had been created during gluster installation in the system itself. But the contents have been created during the test runs, hence clearing them out.

srijan-sivakumar: Let me verify it once.

srijan-sivakumar: @nik-redhat You were right. No issue with clearing out things inside `/var/run/gluster/`. Hence modified the command.

common/ops/support_ops/machine_ops.py
```diff
@@ -145,7 +145,8 @@ def hard_terminate(self, server_list: list, client_list: list,
         # directory on the servers.
         cmd = ("rm -rf /var/lib/glusterd/vols/*; rm -rf /var/lib/glusterd"
                "/peers/*; rm -rf /var/lib/glusterd/snaps/*; "
-               "rm -rf /var/lib/glusterd/glusterfind/.keys/*")
+               "rm -rf /var/lib/glusterd/glusterfind/.keys/*; "
+               "rm -rf /var/run/gluster/*; ")
```
nik-redhat, line 149: It should be `gluster*` I think, else the socket file won't be cleared out unless that is auto-deleted on stopping glusterd?

srijan-sivakumar: No no..everything is cleared out. I verified after a test run. The contents of `/var/run/gluster` is empty.

nik-redhat (approved):

PR #927, opened 2021-08-04
Migrated TC test_default_granular_entry_heal

nik-redhat:
Migrated TC [test_default_granular_entry_heal](https://github.com/gluster/glusto-tests/blob/master/tests/functional/afr/test_default_granular_entry_heal.py)

And, added the op `occurences_of_pattern_in_file`.

Updates: #292 

### DIFF
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -121,7 +121,7 @@ class IoOps(AbstractOps):
             list_of_nodes = [list_of_nodes]
 
         if not isinstance(list_of_paths, list):
-            list_of_paths = (list_of_paths.split(" "))
+            list_of_paths = [list_of_paths.split(" ")]
 
         for path in list_of_paths:
             cmd = f"stat {path}"
@@ -985,6 +985,37 @@ class IoOps(AbstractOps):
         if len(ret['msg']) == 0:
             return 1
         return 0
+
+    def occurences_of_pattern_in_file(self, node: str, search_pattern: str,
+                                      filename: str) -> int:
+        """
+        Get the number of occurences of pattern in the file
+
+        Args:
+            node (str): Host on which the command is executed.
+            search_pattern (str): Pattern to be found in the file.
+            filename (str): File in which the pattern is to be validated
+
+        Returns:
+            int: -1 - When the file doesn't exists,
+                  0 - When pattern doesn't exists in the file,
+                number of occurences - When pattern is found.
+        """
+        # Check if the file exists
+        ret = self.path_exists(node, filename)
+        if not ret:
+            self.logger.error(f"File {filename} is not present on the node")
+            return -1
+
+        cmd = f"grep -c '{search_pattern}' {filename}"
+        ret = self.execute_abstract_op_node(cmd, node)
+        num = int(ret['msg'][0].strip())
+        if num == 0:
+            self.logger.error("No occurence of the pattern found in the file"
+                              f" {filename}")
+            return 0
+
+        return num
 
     def find_and_replace_in_file(self, node: str,
                                  fpattern: str, rpattern: str,
--- a/common/ops/support_ops/io_ops.py
+++ b/common/ops/support_ops/io_ops.py
@@ -970,8 +970,9 @@ class IoOps(AbstractOps):
         ret = self.execute_abstract_op_node(cmd, node)
         return ret['msg']
 
-    def check_if_pattern_in_file(self, node: str,
-                                 pattern: str, fqpath: str):
+    def check_if_pattern_in_file(self, node: str, pattern: str,
+                                 fqpath: str,
+                                 check_occurence: bool = False) -> int:
         """Check if a give pattern is in seen in file or not.
 
         Args:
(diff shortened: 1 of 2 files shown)


nik-redhat:
```
Results for afr
Test nature : disruptive
TestName : test_default_granular_entry_heal
|     rep     |     PASS    |        00:02:29       |      NA     |
|   dist-rep  |     PASS    |        00:00:33       |      NA     |
|     arb     |     PASS    |        00:00:29       |      NA     |
|   dist-arb  |     PASS    |        00:00:35       |      NA     |
|              Tests Ran               |   4   |
Total Time Taken : 00 hours 04 minutes 21 seconds
✔ Tearing down successful.
```

common/ops/support_ops/io_ops.py
```diff
@@ -986,6 +992,37 @@ def check_if_pattern_in_file(self, node: str,
             return 1
         return 0
 
+    def occurences_of_pattern_in_file(self, node: str, search_pattern: str,
```
srijan-sivakumar, line 995: Can't we re-use or re-struturize the function `check_if_pattern_in_file` to be called inside this function or maybe with a simple flag, instead of returning 0 for pattern present in file, if the flag is enabled we can return the number of occurrences. I feel we are just duplicating the logic in this new function `occurences_of_pattern_in_file`.

tests/functional/afr/test_default_granular_entry_heal.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        init_val = arequal[0][-1].split(':')[-1]
+        arequals = self.redant.collect_bricks_arequal(brick_list)
+        for brick_arequal in arequals:
+            brick_total = brick_arequal[-1].split(':')[-1]
+            if init_val != brick_total:
+                raise Exception("Arequals not matching")
+
+    @staticmethod
+    def _add_dir_path_to_brick_list(brick_list):
```
srijan-sivakumar, line 90: Just curious, any specific reason for keeping this method as static method ?

tests/functional/afr/test_default_granular_entry_heal.py
```diff
@@ -0,0 +1,211 @@
...
+            work_dir = f"{self.mountpoint}/mydir"
+            arequals = self.redant.collect_mounts_arequal(self.mounts,
+                                                          work_dir)
+
+            # Get arequal on bricks and compare with mount_point_total
+            self._check_arequal_on_bricks_with_a_specific_arequal(
+                arequals, dir_brick_list)
+
+        # Check arequals for "distributed-replicated"
+        if self.volume_type == "dist-rep" or self.volume_type == "dist-arb":
```
srijan-sivakumar, line 121: Seeing that we just execute the test for rep, arb, dist-rep and dist-arb, we can simply put this check as an else, as the previous if conditional already checks for rep and arb and the else part will cover for dist-rep and dist-arb.

srijan-sivakumar (changes requested):

tests/functional/afr/test_default_granular_entry_heal.py
nik-redhat: Not sure, I just kept it the same as the glusto TC. It doesn't make any difference if we don't make it static. What do you suggest?

tests/functional/afr/test_default_granular_entry_heal.py
nik-redhat: Ack

tests/functional/afr/test_default_granular_entry_heal.py
srijan-sivakumar: Well, in python static functions don't get to use the class's context or the object's context, and most of the times they're used to execute some logic which is independent of object. Seeing the function logic, we are just operating over the parameters being passed and hence I don't see any harm in making it static, I think we should actually mark such methods as static consciously wherein the method just takes in some input parameters, operates over them and returns a response. We can prevent the class or objects' context coming into picture when the interpreter might push this method to the stack, because in case of normal methods and class methods, these functions get the object or class reference as the first parameter by default.

common/ops/support_ops/io_ops.py
nik-redhat: So you mean we should have only one function like `check_pattern_in_file` and use flags to do the counting as well? Or, have both but reuse the other one inside this one?

srijan-sivakumar: I feel we can just go with the first method. Just use check_pattern_in_file and with the flag enabled, return the number of occurrences.

nik-redhat: I think if we keep one function then there will be a lot of nested `if-else` which will necessarily make the code complex. Using the other function inside this function seems better to me.

nik-redhat: Anyways, let me check and see how it goes, I will push the changes once done.

srijan-sivakumar: Do check it out...and see which is better...

nik-redhat: I have kept the new function and reused the other function due to issues with return value and its usage.

nik-redhat:
```
Results for afr
Test nature : disruptive
TestName : test_default_granular_entry_heal
|     rep     |     PASS    |        00:00:35       |      NA     |
|   dist-rep  |     PASS    |        00:00:40       |      NA     |
|     arb     |     PASS    |        00:00:32       |      NA     |
|   dist-arb  |     PASS    |        00:00:39       |      NA     |
|              Tests Ran               |   4   |
Total Time Taken : 00 hours 02 minutes 42 seconds
✔ Tearing down successful.
```

Sheetalpamecha (approved):

srijan-sivakumar (approved):

Issue #924, opened 2021-08-03
Doc to help decide test nature


srijan-sivakumar:
There's a need for a doc containing the points to be considered for determining the test nature as disruptive or non disruptive.

PR #930, opened 2021-08-08
[Doc] Deciding test nature markdown added

srijan-sivakumar:
### All Submissions:

Description: The added markdown helps people to understand what points are to be kept in mind for a test case to be marked as disruptive/non-disruptive.

Fixes: #924

### DIFF
--- a/docs/BP/TestNature.md
+++ b/docs/BP/TestNature.md
@@ -4,7 +4,7 @@ Redant test cases can be classified into two types,
 1. Disruptive test cases (hereon referred to as Dtc)
 2. Non-Disruptive test cases (hereon referred to as NDtc)
 
-Based on the design of redant, NDtcs can be run paralelly while Dtcs have to be run sequentially. So, how does one decide whether the test case they're writing is a NDtc or a Dtc ?
+Based on the design of redant, NDtcs are run paralelly while Dtcs are run sequentially. So, how does one decide whether the test case they're writing is a NDtc or a Dtc ?
 
 Following are some pointers which describe a Dtc. If a test case satisfies even one of the stated points, it is Dtc. For a test case to be NDtc, it shouldn't satisfy any of the stated points.
 
--- a/docs/BP/TestNature.md
+++ b/docs/BP/TestNature.md
@@ -1,0 +1,14 @@
+## How to decide the test nature
+
+Redant test cases can be classified into two types,
+1. Disruptive test cases (hereon referred to as Dtc)
+2. Non-Disruptive test cases (hereon referred to as NDtc)
+
+Based on the design of redant, NDtcs can be run paralelly while Dtcs have to be run sequentially. So, how does one decide whether the test case they're writing is a NDtc or a Dtc ?
+
+Following are some pointers which describe a Dtc. If a test case satisfies even one of the stated points, it is Dtc. For a test case to be NDtc, it shouldn't satisfy any of the stated points.
+
+* The test case reboots a node ( server or client ).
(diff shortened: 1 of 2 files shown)


docs/BP/TestNature.md
```diff
@@ -0,0 +1,14 @@
...
+2. Non-Disruptive test cases (hereon referred to as NDtc)
+
+Based on the design of redant, NDtcs are run paralelly while Dtcs are run sequentially. So, how does one decide whether the test case they're writing is a NDtc or a Dtc ?
+
+Following are some pointers which describe a Dtc. If a test case satisfies even one of the stated points, it is Dtc. For a test case to be NDtc, it shouldn't satisfy any of the stated points.
+
+* The test case reboots a node ( server or client ).
+* The test case reconfigures ports or changes firewalld rules.
+* The test case is unmounting the mountpoint which acts as glusterfs brick roots.
+* The test case aims to create a huge amount of volumes, snapshots, geo-rep sessions or gfind sessions.
```
nik-redhat, line 14: Also:
- The test case performs any cluster-level operations

nik-redhat, line 14: And, consumes high memory or performs huge IO (can be written in a better way :p)

docs/BP/TestNature.md
```diff
@@ -0,0 +1,14 @@
...
+1. Disruptive test cases (hereon referred to as Dtc)
+2. Non-Disruptive test cases (hereon referred to as NDtc)
+
+Based on the design of redant, NDtcs are run paralelly while Dtcs are run sequentially. So, how does one decide whether the test case they're writing is a NDtc or a Dtc ?
+
+Following are some pointers which describe a Dtc. If a test case satisfies even one of the stated points, it is Dtc. For a test case to be NDtc, it shouldn't satisfy any of the stated points.
+
+* The test case reboots a node ( server or client ).
+* The test case reconfigures ports or changes firewalld rules.
+* The test case is unmounting the mountpoint which acts as glusterfs brick roots.
```
nik-redhat, line 13: mountpoint or the volume bricks itself.

nik-redhat (changes requested):

docs/BP/TestNature.md
srijan-sivakumar: Ack

docs/BP/TestNature.md
srijan-sivakumar: Ack

docs/BP/TestNature.md
srijan-sivakumar: Hmm..this is a little vague. Even if we state consumes high memory, the question would be what is the value we consider as high memory and huge IO. If someone with say an env of 3 servers with 2 GB RAM, a disk per server for brick and 2 CPU cores were to run then the definition will change. I guess for now we can prop up volume count and other counts as a parameter for high memory consumption, till we actually benchmark and find a value which is the actual limitation..

srijan-sivakumar: So, maybe we can just provide the example of our env along with the volume counts till which we consider it non disruptive. Any pointers on till what count have you tried out non disruptive tests ?

nik-redhat: Make sense :thinking: ...... what I have seen is that we can get a maximum of 75 distribute (x3) volumes in a 3 server env having 2GB RAM and 2 CPUs. But again, adding such data won't make much sense either, so I think guess the current point can be kept as it is for the time being.

nik-redhat (approved):

Issue #932, opened 2021-08-09
Log addition to env and result handler


srijan-sivakumar:
The result_handler, env_seutp and env_teardown component requires logging so as to track any errors or failures happening.

PR #933, opened 2021-08-09
[Core] Increased debug log penetration.

srijan-sivakumar:
Increased the amount of debug logs in redant_main
and also put some basic logging inside the result
handler.

Fixes: #932

### DIFF
--- a/core/environ.py
+++ b/core/environ.py
@@ -136,10 +136,11 @@ class environ:
             self.redant.create_cluster(self.server_list)
             self._check_and_copy_io_script()
             self._check_and_install_arequal_checksum()
+            self.redant.logger.info("Environment setup success.")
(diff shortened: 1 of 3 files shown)


core/environ.py
```diff
@@ -153,10 +154,11 @@ def teardown_env(self):
         try:
```
nik-redhat, line 154: A log before calling this function like, `Starting environment teardown` or something of that sort?

nik-redhat: And a similar one for the setup part as well?

srijan-sivakumar: For setup I've added a debug log in the redant_main itself. I'll add a similar log for teardown as well.

nik-redhat (approved):

Issue #938, opened 2021-08-09
Argument order wrong in result handler


srijan-sivakumar:
Following is the traceback received due to the wrong order of arguments passed.

```python
Traceback (most recent call last):
  File "./core/redant_main.py", line 151, in <module>
    main()
  File "./core/redant_main.py", line 142, in main
    handle_results(result_queue, total_time, args.excel_sheet,
  File "/root/redant-test/redant/core/result_handler.py", line 384, in handle_results
    logger.debug("Initializing result handling.")
AttributeError: 'str' object has no attribute 'debug'
```

PR #939, opened 2021-08-09
[Core] Fixing argument order

srijan-sivakumar:
The logger_obj has to be the 3rd argument instead of 4th argument.

Fixes: #938


srijan-sivakumar:
After this change, the result handler goes ahead and creates the file.

nik-redhat (approved):

Issue #900, opened 2021-07-28
Revisiting lint exclusion


srijan-sivakumar:
We need to revisit the lint checks we have excluded to verify if we have ignored any valid case.

srijan-sivakumar:
The flake ignores the following rules 
C901,F811,F841,E722,E226,E402,W503,W605

And in the pylintrc we have the following in excluded state:
        parameter-unpacking,
        unpacking-in-except,
        old-raise-syntax,
        backtick,
        long-suffix,
        old-ne-operator,
        old-octal-literal,
        import-star-module-level,
        non-ascii-bytes-literal,
        raw-checker-failed,
        bad-inline-option,
        locally-disabled,
        locally-enabled,
        file-ignored,
        suppressed-message,
        useless-suppression,
        deprecated-pragma,
        apply-builtin,
        basestring-builtin,
        buffer-builtin,
        cmp-builtin,
        coerce-builtin,
        execfile-builtin,
        file-builtin,
        long-builtin,
        raw_input-builtin,
        reduce-builtin,
        standarderror-builtin,
        unicode-builtin,
        xrange-builtin,
        coerce-method,
        delslice-method,
        getslice-method,
        setslice-method,
        no-absolute-import,
        old-division,
        dict-iter-method,
        dict-view-method,
        next-method-called,
        metaclass-assignment,
        indexing-exception,
        raising-string,
        reload-builtin,
        oct-method,
        hex-method,
        nonzero-method,
        cmp-method,
        input-builtin,
        round-builtin,
        intern-builtin,
        unichr-builtin,
        map-builtin-not-iterating,
        zip-builtin-not-iterating,
        range-builtin-not-iterating,
        filter-builtin-not-iterating,
        using-cmp-argument,
        eq-without-hash,
        div-method,
        idiv-method,
        rdiv-method,
        exception-message-attribute,
        invalid-str-codec,
        sys-max-int,
        bad-python3-import,
        deprecated-string-function,
        deprecated-str-translate-call,
        deprecated-itertools-function,
        deprecated-types-field,
        next-method-defined,
        dict-items-not-iterating,
        dict-keys-not-iterating,
        dict-values-not-iterating,
        attribute-defined-outside-init,
        missing-docstring,
        E1101,
        E265,
        E0401,
        R0801,
        R0903,
        C0103,
        W0703,
        R0902,
        W0511,
        R0913,
        R0914,
        R1702,
        R0912,
        E1121,
        E0102,
        R1720,
        R0201,
[...]

PR #944, opened 2021-08-10
[Tool] Linting behavior similar to GH Actions

srijan-sivakumar:
The linting check needs to be similar to GH actions,
plus a little simple.

Updates: #900

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,7 +32,7 @@ To execute the same:
 ### Using script:
 You can run the following command
 ```console
-tools/linting.sh -p PATH/TO/FILE -l -f
+tools/linting.sh
 ```
 To know more about the script, go [here](./tools/README.md).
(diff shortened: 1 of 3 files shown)


nik-redhat:
Why was then the linting script different earlier?

srijan-sivakumar:
> Why was then the linting script different earlier?

We had created the linting script first and then setup the GH actions and modified the values in GH actions' exclusion list. Plus, the linting script was too complex. We can simply run the linting script over the complete code..and if at all there's any issue, we'd get it. We can simply use this linting script before every code push so as to prevent the repeated pushes we have to do sometimes for lint fixes.

nik-redhat:
Well, I know it was complex but I guess that was made that way (intentionally). Anyways, I think we are fine without that added complexity.

nik-redhat (approved):

PR #946, opened 2021-08-11
Migrated TC test_heal_for_conservative_merge_with_two_bricks_blame.py

Sheetalpamecha:
Migrated TC [test_heal_for_conservative_merge_with_two_bricks_blame.py](https://github.com/gluster/glusto-tests/blob/HEAD/tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py)

Updates: #292

### DIFF
--- a/tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py
+++ b/tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py
@@ -1,82 +1,33 @@
-#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
-#
-#  This program is free software; you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
-#  the Free Software Foundation; either version 2 of the License, or
-#  any later version.
-#
-#  This program is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
-#
-#  You should have received a copy of the GNU General Public License along
-#  with this program; if not, write to the Free Software Foundation, Inc.,
-#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+"""
+ Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+
+ This program is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation; either version 2 of the License, or
+ any later version.
+
+ This program is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License along
+ with this program; if not, write to the Free Software Foundation, Inc.,
+ 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+Description:
+    Test for conservative merge between two bricks.
+"""
+
+# disruptive;rep
 
 from time import sleep
-
-from glusto.core import Glusto as g
-from glustolibs.gluster.brick_libs import (get_all_bricks, are_bricks_offline,
-                                           bring_bricks_offline,
-                                           get_online_bricks_list,
-                                           are_bricks_online)
-from glustolibs.gluster.exceptions import ExecutionError
-from glustolibs.gluster.gluster_base_class import GlusterBaseClass, runs_on
-from glustolibs.gluster.glusterdir import mkdir
-from glustolibs.gluster.gluster_init import restart_glusterd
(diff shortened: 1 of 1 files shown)


Sheetalpamecha:
```
Results for heal
Test nature : disruptive
TestName : test_heal_for_conservative_merge_with_two_bricks_blame
|     rep     |     PASS    |        00:01:33       |      NA     |
Total Time Taken : 00 hours 02 minutes 04 seconds
✔ Tearing down successful.
```

tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # Reset the xattrs from brick3 on to brick2
+        second_xattr_to_reset = f"trusted.afr.{self.vol_name}-client-1"
+        redant.set_fattr(f'{brick_path}/dir1', second_xattr_to_reset,
+                         host, xattr_value)
+
+        # Bring brick2 online
+        redant.restart_glusterd(self.server_list[1])
+
+        sleep(10)
```
srijan-sivakumar, line 119: Instead of sleep, I think the better way would be to check,
1. If glusterd has started, there is this method called, [`wait_for_glusterd_to_start`](https://github.com/gluster/redant/blob/0e8fa1640127a716b9d3788b7e647acc598e04c5/common/ops/gluster_ops/glusterd_ops.py#L213)
2. Once we are sure of glusterd status, the next validation we can put is to check if peers are connected using [`wait_for_peers_to_connect`](https://github.com/gluster/redant/blob/0e8fa1640127a716b9d3788b7e647acc598e04c5/common/ops/gluster_ops/peer_ops.py#L378)
3. We need to check if brick process is started in node 2 using [`wait_for_bricks_to_come_online`](https://github.com/gluster/redant/blob/0e8fa1640127a716b9d3788b7e647acc598e04c5/common/ops/gluster_ops/brick_ops.py#L922) and herein we can pass in only brick1 as part of the brick list.

srijan-sivakumar: with these checks in place we can actually remove the check we've put for brick2 using `are_bricks_online`.

srijan-sivakumar (changes requested):

tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py
Sheetalpamecha: Yes, makes sense. Its indeed a better way. I will make the changes

nik-redhat: The `wait_for_peers_to_connect` is not needed, in fact the `wait_for_glusterd_to_start` is also not needed as we are already doing a check for bricks in a loop for 100secs, so if at all glusterd failed to start the `wait_for_bricks_to_come_online` will report the error,

srijan-sivakumar: 🤔 true, but wanted the right function to report the error if it occurs. Hence the suggestion. But we can try using`wait_for_bricks_to_come_online` directly, I think that will handle these cases accordingly.

nik-redhat: Just trying to avoid unwanted checks and sleep, but if need be at max we should use the `wait_for_glusterd_to_start` function and not the peer's function.

Sheetalpamecha: okay, Finally went ahead with ```wait_for_glusterd_to_start```  and ```wait_for_bricks_to_come_online``` function and not the peer's function.

Sheetalpamecha:
```Results for heal
Test nature : disruptive
TestName : test_heal_for_conservative_merge_with_two_bricks_blame
|     rep     |     PASS    |        00:01:35       |      NA     |
Total Time Taken : 00 hours 02 minutes 06 seconds
✔ Tearing down successful.```

File: tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py [hunk 2]
```diff
@@ -0,0 +1,142 @@
...
+
+        # Reset the xattrs from brick3 on to brick2
+        second_xattr_to_reset = f"trusted.afr.{self.vol_name}-client-1"
+        redant.set_fattr(f'{brick_path}/dir1', second_xattr_to_reset,
+                         host, xattr_value)
+
+        # Bring brick2 online
+        redant.restart_glusterd(self.server_list[1])
+
+        if not (self.redant.
```
nik-redhat, line 119: nit: it should be only `redant`

srijan-sivakumar (approved):

tests/functional/afr/heal/test_heal_for_conservative_merge_with_two_bricks_blame.py
srijan-sivakumar: I guess we can ignore this nit as `self.redant` and `redant` is the same variable. But it'd be advisable to take `redant` as standard nomenclature when calling methods in the test case.

Issue #968, opened 2021-08-18
Restructurizing Docs


srijan-sivakumar:
Adding to the modifications already done by @nik-redhat , further changes to doc layout.

PR #969, opened 2021-08-18
[Doc] Updating doc structure

srijan-sivakumar:
Restructuring Doc dir structure.

Namely,
1. Brought `Ops` under `Common`.
2. Merged `volds.md` and `environ.md`
3. Removed `rexe.md` under `Common` as it didn't add much value.
4. Added updated docs index link in main `README.md`
5. Added a link in doc index to flow back to main menu.
6. Changed the naming convention of doc files to snake case.

Fixes: #968


docs/Common/README.md
```diff
@@ -3,3 +3,5 @@
 ## Index
```
nik-redhat, line 3: This header can be removed, as we have a `Common Index` header just above it.

docs/README.md
```diff
@@ -2,10 +2,10 @@
 
 * [Common](./Common)
 * [Core](./Core)
-* [Ops](./Ops)
+* [Ops Index](./Common/Ops)
```
nik-redhat, line 5: I think this can be removed, and we should only keep the items that are at the same level in this Index.

nik-redhat: Also, a link to `code_considerations` can also be added. ( I forgot to add :p)

Sheetalpamecha:
Apart from nikhil's comment, LGTM

nik-redhat (approved):

PR #992, opened 2021-08-30
Migrated TC test_self_heal_daemon_process.py

Sheetalpamecha:
[Migrated TC test_self_heal_daemon_process.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/afr/heal/test_self_heal_daemon_process.py)

Updates: #292


Sheetalpamecha:
This func _test_brick_process_not_started_on_read_only_node_disks is failing in this TC. Tried it in glusto as well. Its failing there too.

nik-redhat (approved):
Yeah, the read-only mount part doesn't work anymore.

PR #993, opened 2021-08-30
Migrated TC test_self_heal.py

Sheetalpamecha:
[Migrated TC test_self_heal.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/afr/heal/test_self_heal.py)

Updates: #292


Sheetalpamecha:
```
✖ test_self_heal-rep Failed
Results for heal
Test nature : disruptive
TestName : test_self_heal
|     rep     |     FAIL    |        01:37:29       |      NA     |
|   Disruptive Tests Pass Percentage   |  0.0  |
|        Total Pass Percentage         |  0.0  |
Total Time Taken : 01 hours 38 minutes 07 seconds
```

nik-redhat:
I have tested the TC multiple times on the glusto shared env and the TC works fine on that. Though, there are spurious failures sometimes.
```
Results for heal
Test nature : disruptive
TestName : test_self_heal
|     rep     |     PASS    |        00:58:40       |      NA     |
Total Time Taken : 00 hours 58 minutes 51 seconds
✔ Tearing down successful.
```

Sheetalpamecha:
Yes I have tried on glusto too and it passes sometimes and sometimes heal timed out.
Can you check on redant setup too?  I didn't have any luck in getting it run successfully in my redant setup

nik-redhat:
Yeah, that's the output using Redant only.

Sheetalpamecha:
> Yeah, that's the output using Redant only.
yeah overlooked it
> The only failure I see is during IO validation, due to the transport endpoint not being connected, and nothing else.
I also have seen IO validation error couple of times but mostly with heal not completiong

We can take it in without adding in excluding TCs but there is chance of it failing spuriously with either error

nik-redhat:
Now, I have tested on my env as well, and the TC passes there too.
Just one thing, that is we require a bit better VMs than what we use locally because the issue that we are facing in the TC is that while writing on the client we are also bringing brick down and then up one by one. And, since the VMs are not so good the execution of the IO process takes a bit of time and while doing the IO we have 2 bricks down at the same time (or at least before the client-side quorum could be satisfied) and as a reason we are getting `Transport endpoint not connected` error, but if we add a `sleep` of say 5 secs then the TC is working fine.

Sheetalpamecha:
Tried with sleep 5. It might prevent Io validation error but still heal not completing issue may come during runs
```
✖ test_self_heal-rep Failed
Results for heal
Test nature : disruptive
TestName : test_self_heal
|     rep     |     FAIL    |        00:51:05       |      NA     |
|   Disruptive Tests Pass Percentage   |  0.0  |
|        Total Pass Percentage         |  0.0  |
Total Time Taken : 00 hours 51 minutes 45 seconds

[2021-09-01 14:42:00,857] ERROR  - Heal has not yet completed
[2021-09-01 14:42:00,857] ERROR  - Traceback (most recent call last):
  File "/home/user/redant/./tests/d_parent_test.py", line 87, in parent_run_test
    self.run_test(self.redant)
  File "/home/user/redant/./tests/functional/afr/heal/test_self_heal.py", line 370, in run_test
    self._test_self_heal_50k_files_heal_default()
  File "/home/user/redant/./tests/functional/afr/heal/test_self_heal.py", line 340, in _test_self_heal_50k_files_heal_default
    raise Exception("Heal has not yet completed")
Exception: Heal has not yet completed
```

nik-redhat:
What is the size of the cluster that you have?

Sheetalpamecha:
5GB

> What is the size of the cluster that you have?
> Reason being with sleep the TC passes for me always sweat_smile

5GB

nik-redhat:
I mean how many servers and clients do you have and is any server being used as a client as well?

Sheetalpamecha:
okay. 3 servers and 1 client. and 1 node is used as both server and client

nik-redhat:
Ok, I doubt this might come into play..... but since we have contradicting results I think we can let the TC run in nightly CI and check the results then we can take the action accordingly.
But, do add the sleep at the end of the loop in the TC.

tests/functional/afr/heal/test_self_heal.py
```diff
@@ -0,0 +1,371 @@
...
+                                    f"{mount_obj['client']}:"
+                                    f"{mount_obj['mountpath']}")
+            proc = self.redant.create_files(num_files=50000,
+                                            fix_fil_size="1k",
+                                            path=mount_obj['mountpath'],
+                                            node=mount_obj['client'])
+            self.all_mounts_procs.append(proc)
+        self.io_validation_complete = False
+
+        # Validate IO
```
Sheetalpamecha, line 305: Here right? After creation of all files

tests/functional/afr/heal/test_self_heal.py
```diff
@@ -0,0 +1,371 @@
...
+            # Verify volume's all processes are online
+            if not (self.redant.verify_all_process_of_volume_are_online(
+                    self.vol_name, self.server_list[0])):
+                raise Exception("All process of volume are not online")
+
+            # Wait for self-heal-daemons to be online
+            if not self.redant.is_shd_daemonized(self.server_list):
+                raise Exception("Self Heal Daemon process was still"
+                                " holding parent process.")
+
```
nik-redhat, line 252: No no...here inside the loop.

Sheetalpamecha: But this is `_test_data_self_heal_command` func. which is not causing any problem.

`_test_self_heal_50k_files_heal_default `when healing 50k files the problem is occuring

nik-redhat: Strange :joy: .... this is where I was seeing failures previously and no issues with the second test.

Sheetalpamecha: okay let me remove this from excluded tests and let redant-nightly build run it

nik-redhat (approved):
Let's take it in without sleep for now, and check the results for 2-3 runs if the test fails for heal we will mark it as excluded but if it fails in IO validation, we can add sleep and try again.

PR #994, opened 2021-08-30
Migrated TC test_write_io_mount_point_resumed_quorum_restored

nik-redhat:
Migrated TC [test_write_io_mount_point_resumed_quorum_restored](https://github.com/gluster/glusto-tests/blob/master/tests/functional/afr/test_write_io_mount_point_resumed_quorum_restored.py)

Updates: #292 

### DIFF
--- a/common/ops/support_ops/machine_ops.py
+++ b/common/ops/support_ops/machine_ops.py
@@ -71,16 +71,27 @@ class MachineOps(AbstractOps):
         """
         Wait for a node to come up online.
         Arg:
-            node (str)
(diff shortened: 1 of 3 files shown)


nik-redhat:
This TC has been marked as excluded because it depends on the time taken to reboot a system, and in the meantime, IO is performed. And, as a result, IO passes (though it should fail) as the reboot was already done.
The TC is marked as excluded for glusto as well.

Sheetalpamecha (approved):

Sheetalpamecha (approved):

PR #997, opened 2021-08-30
Migrated TC test_256_snapshots

nik-redhat:
Migrated TC [test_256_snapshots](https://github.com/gluster/glusto-tests/blob/master/tests/functional/snapshot/test_256_snapshots.py)

Updates: #292 

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -66,4 +66,4 @@ excluded_tests:
     - tests/functional/afr/test_arb_to_repl_conversion_with_io.py # Heal not completing even after increasing the timeout and the IO is completed
     - tests/functional/afr/heal/test_self_heal_daemon_process.py # TC is checking for conditions that doesn't happen in glusterfs. It is failing in glusto too
     - tests/functional/afr/test_write_io_mount_point_resumed_quorum_restored.py # Probability of the TC passing is very less as the TC depends on the time of reboot, which varies
-    - tests/functional/snapshot/test_256_snapshots.py # Requires baremetal servers with big disks to create these many snapshots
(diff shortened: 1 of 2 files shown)


nik-redhat:
@srijan-sivakumar for this TC I doubt we will have the hardware to create 256 snapshots? Shall I add it to the excluded list?

srijan-sivakumar:
For this TC, we would need a baremetal server or a VM with good amount of resource. Both of which are not with us currently. I think this again needs to be put into the exclude list with the proper description/reasoning.

nik-redhat:
> For this TC, we would need a baremetal server or a VM with good amount of resource. Both of which are not with us currently. I think this again needs to be put into the exclude list with the proper description/reasoning.

Yeah, made the update.

config/config.yml
```diff
@@ -66,3 +66,4 @@ excluded_tests:
     - tests/functional/afr/test_arb_to_repl_conversion_with_io.py # Heal not completing even after increasing the timeout and the IO is completed
     - tests/functional/afr/heal/test_self_heal_daemon_process.py # TC is checking for conditions that doesn't happen in glusterfs. It is failing in glusto too
     - tests/functional/afr/test_write_io_mount_point_resumed_quorum_restored.py # Probability of the TC passing is very less as the TC depends on the time of reboot, which varies
+    - tests/functional/snapshot/test_256_snapshots.py # Requires baremetal servers with big disks to create these many snapshots
```
srijan-sivakumar, line 69: nit: Rather than saying big disk, we can state something like multiple disks and good amount of RAM.

srijan-sivakumar (changes requested):

config/config.yml
nik-redhat: Done

srijan-sivakumar (approved):

Issue #1000, opened 2021-08-30
test_metadata_self_heal_on_open throwing unexpected traceback on failure


srijan-sivakumar:
[2021-08-30 19:13:40,006] INFO  - Running test_metadata_self_heal_on_open_fd-rep
[2021-08-30 19:13:40,970] ERROR  - test_metadata_self_heal_on_open_fd-rep : 'TestCase' object has no attribute 'nodes'
[2021-08-30 19:13:40,970] ERROR  - test_metadata_self_heal_on_open_fd-rep : Traceback (most recent call last):
  File "/root/redant-test/redant/core/runner_thread.py", line 49, in run_thread
    self.terminate_test_func()
  File "./tests/functional/arbiter/test_metadata_self_heal_on_open_fd.py", line 34, in terminate
    for node in self.nodes:
AttributeError: 'TestCase' object has no attribute 'nodes'

srijan-sivakumar:
@nik-redhat  this is the scenario when we define certain flag in the beginning of the test case but then the parent_run_test itself fails. Then we directly go to terminate without even defining the flag and hence the terminate also fails.

nik-redhat:
Correct that's the reason I never keep anything outside the `try` block in the terminate. But, I guess this PR sneaked upon when I reviewed it :thinking: ..... anyways I think the way we are handling variables right now is totally fine. Just that we need to be a bit cautious while reviewing, in case someone misses out.

srijan-sivakumar:
Cool. Then I'll push the changes by putting it inside the try-except block.

PR #1003, opened 2021-08-30
[Test] Bringing custom terminate under t-e block

srijan-sivakumar:
Following changes have been done in this patch,
1. Bringing the custom terminate logic completely
under the try-except block as the check flag/variable
might not even be defined before its access.
2. Modified how the self.nodes is formed to
restrict the inclusion of a node more than once in it.
This situation can arise if we were to use the same
node as both server and client.

Note: The test case is failing due to mismatch
in username expected. But this patch is not
a fix for that.

Fixes: #1000


srijan-sivakumar:
Issue opened for test failure : [1004](https://github.com/gluster/redant/issues/1004). This issue can be fixed separately.

Sheetalpamecha (approved):

PR #1045, opened 2021-09-07
Fix f-string and object name

nik-redhat:
**Description:**

Fixed `f-string` issue for a few TCs.
Updated the object name to be `redant` in one TC
Marked a TC as excluded as it works only in downstream RHGS code


nik-redhat:
Not attaching results as these are very minute changes.

Sheetalpamecha (approved):

PR #1053, opened 2021-09-08
Use deepcopy() method for copying dict

nik-redhat:
## Description:

In python, when we copy a dictionary it creates a new reference to the old dictionary, and thereby if we update the values in
the new (reference) dictionary, then the old one is also affected. The same has been the case in Redant for the `volume_config` that is being used while creating volumes.
The value of the `self.vol_type_inf` dictionary was copied using the `=` operator and then the contents of the new reference/copy were updated as a result the actual dict was also getting updated and caused spurious failures in the CI to run.

**Fix:**
To overcome, this issue it is recommended to use the  `deepcopy()` method in python to copy such dicts, lists so that instead of a new reference, a completely new object is created from the original reference.

### DIFF
--- a/tests/functional/afr/heal/test_data_split_brain_resolution.py
+++ b/tests/functional/afr/heal/test_data_split_brain_resolution.py
@@ -33,7 +33,7 @@ class TestCase(DParentTest):
         """
         Override the volume create
         """
-        conf_hash = self.vol_type_inf['rep']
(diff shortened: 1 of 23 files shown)


nik-redhat:
Ready for review.

srijan-sivakumar (approved):

PR #1094, opened 2021-09-20
Migrate TC test_wipe_out_directory_permissions.py

Sheetalpamecha:
Migrate TC [test_wipe_out_directory_permissions.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/dht/test_wipe_out_directory_permissions.py)

Updates: #292 

### DIFF
--- a/tests/functional/dht/test_wipe_out_directory_permissions.py
+++ b/tests/functional/dht/test_wipe_out_directory_permissions.py
@@ -1,0 +1,132 @@
+#  Copyright (C) 2020 Red Hat, Inc. <http://www.redhat.com>
+#
+#  This program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation; either version 2 of the License, or
+#  any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License along`
+#  with this program; if not, write to the Free Software Foundation, Inc.,
+#  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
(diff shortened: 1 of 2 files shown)


Sheetalpamecha:
```
Results for dht
Test nature : disruptive
TestName : test_wipe_out_directory_permissions
|     dist    |     PASS    |        00:01:54       |      NA     |
Total Time Taken : 00 hours 03 minutes 56 seconds

```

tests/functional/dht/test_wipe_out_directory_permissions.py
```diff
@@ -0,0 +1,122 @@
...
+    def _check_permissions_of_dir(self):
+        """Check permissions of dir created."""
+        bricks = self.redant.get_all_bricks(self.vol_name,
+                                            self.server_list[0])
+
+        for brick_path in bricks:
+            node, path = brick_path.split(":")
+            ret = self.redant.get_file_stat(node, "{}/dir".format(path))
+            if not ret:
+                raise Exception("Not able to get stat of the dir")
```
nik-redhat, line 52: We don't seem to be validating the expected permissions here.

nik-redhat (changes requested):

tests/functional/dht/test_wipe_out_directory_permissions.py
Sheetalpamecha: Added the check. But there seems to be a bug here. Directory permissions changes on cd/ls operation on mountpoint without explicit changing from us. Tried it manually too and ran into this bug.

nik-redhat: Ok, so did they have some other permission?

Sheetalpamecha: No

nik-redhat: Ok, it seems to be working on downstream code but not on upstream latest code.
Then, I think it is better to mark it as excluded with this justification.

Sheetalpamecha: okay

nik-redhat (approved):

PR #1191, opened 2021-10-20
Migrate TC test_basic_memory_leaks.py 

Sheetalpamecha:
Migrate TC [test_basic_memory_leaks.py](https://github.com/gluster/glusto-tests/blob/master/tests/functional/resource_leak/test_basic_memory_leaks.py)

Updates: #292
Signed-off-by: Sheetal Pamecha spamecha@redhat.com


nik-redhat:
> @nik-redhat Can you run this TC also on CI?
> 
> My setup does not have enough resource, We are doing heavy io's and shared env is in use till friday. Also its taking ~3-4 hrs for a each volume

Sure, currently #1187 is running. Once, that's done, I will trigger it for this one.

PR #1193, opened 2021-10-20
Fixed column width for excel sheets in result handler

nik-redhat:
### All Submissions:

**Description:**
Currently, when an excel sheet is used to save results after the TC run, the width of columns didn't use to
adjust based on the length of data the cell has. And, as a result, some data was clipped off, and the column width had to be manually increased to view the complete data.

**Fix:**
Updated the code, to check for the length of data in each cell and update the column width based on that.

### DIFF
--- a/core/result_handler.py
+++ b/core/result_handler.py
@@ -235,7 +235,8 @@ def _transform_to_percent(statDict: dict) -> dict:
     return tStatDict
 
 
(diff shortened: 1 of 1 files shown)


nik-redhat:
Attaching results.
[results_redant.xls](https://github.com/gluster/redant/files/7386267/results_redant.xls)
[results_redant_1.xls](https://github.com/gluster/redant/files/7386268/results_redant_1.xls)

PR #1198, opened 2021-11-01
Added sleep in some TCs to compensate for mounting of volume after a restart

nik-redhat:
**Description:**
Marked the TC `test_shared_storage` as excluded and added sleep in some TCs to compensate
for a remount of volume after a restart of bricks for a volume. 
**NOTE:""
It is not ideal to add a sleep for in these TCs, but it has been done to adjust for all types of env and gluster installation.


nik-redhat:
It is not always reproducible in the local clusters but happens more frequently in CI.

Sheetalpamecha (approved):

